PR #92, opened 2017-09-21
Add Code Climate config file

### LINKED ISSUE #88: Fix Code Climate Issues
You can take a look into https://codeclimate.com/github/rootstrap/rails_api_base/issues

- [ ] Disable "Missing frozen string literal comment." in the code climate config file like CC explains in that screen (click exclude)
- [ ] Fix order of gems
- [ ] Fix ActiveAdmin long line

### LINKED ISSUE #91: Add/Define a new code_climate.yml file.
This PR https://github.com/rootstrap/rails_api_base/pull/90 removes the current Code Climate configuration since it was invalid.

We should discuss and define the proper configuration and creating a new `code_climate.yml` file according with what we want.

MaicolBen:
Resolves #88
Resolves #91

- Skipped rubocop in db because it complains `Assignment Branch Condition size for change is too high.` for big  `change` methods in migrations
- Skipped flog in spec/config/db because it complains for complex methods in migrations, specs and initializers
- Skipped duplication in spec because there are cases when the tests makes so dynamic that will be hard to debug

### DIFF
--- a/spec/requests/api/v1/sessions/create_spec.rb
+++ b/spec/requests/api/v1/sessions/create_spec.rb
@@ -1,6 +1,28 @@
 require 'rails_helper'
 
 describe 'POST api/v1/users/sign_in', type: :request do
+  shared_examples 'log in successfully' do
+    it 'returns success' do
+      expect(response).to be_success
+    end
+
+    it 'returns the user' do
+      expect(json[:user][:id]).to eq(user.id)
+      expect(json[:user][:email]).to eq(user.email)
+      expect(json[:user][:username]).to eq(user.username)
+      expect(json[:user][:uid]).to eq(user.uid)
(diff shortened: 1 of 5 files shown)


spec/requests/api/v1/sessions/create_spec.rb
```diff
@@ -1,6 +1,28 @@
 require 'rails_helper'
 
 describe 'POST api/v1/users/sign_in', type: :request do
+  shared_context 'log in successfully' do
```
santiagovidal: I think this should be `shared_examples`.

MaicolBen: Nice, I have been using shared contexts wrong for a long time, thanks!!!

PR #97, opened 2017-10-12
Add region to Carriewave initializer file

matiasmansilla1989:
### Description
Add bucket region setting to `Carriewave` config initializer file.


MaicolBen:
Sorry, I just reminded. Can you add this new env to https://github.com/rootstrap/rails_api_base/blob/master/config/application.yml.example ?

matiasmansilla1989:
@MaicolBen Good catch, I fixed that comment.

PR #98, opened 2017-10-20
Move strings to locals

### LINKED ISSUE #47: Use I18n
Use it in every text that is going to be used. For example, error messaging in `ApiController`

MaicolBen:
Resolve #47

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -24,12 +24,12 @@ module Api
 
       def render_error(exception)
         logger.error(exception) # Report to your error managment tool here
-        render json: { error: 'An error ocurred' }, status: 500 unless performed?
+        render json: { error: I18n.t('api.errors.server') }, status: 500 unless performed?
       end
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/api_controller.rb
```diff
@@ -24,12 +24,12 @@ def status
...
       def render_error(exception)
         logger.error(exception) # Report to your error managment tool here
-        render json: { error: 'An error ocurred' }, status: 500 unless performed?
+        render json: { error: I18n.t('api.errors.500') }, status: 500 unless performed?
       end
 
       def render_not_found(exception)
         logger.info(exception) # for logging
-        render json: { error: "Couldn't find the record" }, status: :not_found
+        render json: { error: I18n.t('api.errors.404') }, status: :not_found
```
matiasmansilla1989: WDYT about `api.errors.not_found`? We are following that approach in the `sessions_controller.rb`, example: `api.facebook.not_authorized`

PR #102, opened 2017-10-26
Rename factory girl

MaicolBen:
This PR modify the `factory_girl_rails` gem  since  from 4.9.0 release the gem will be officially deprecated. More details can be found [here](https://robots.thoughtbot.com/factory_bot)


GAKINDUSTRIES:
Description Updated

PR #103, opened 2017-11-07
Use Attributes and Data structures in Api Blueprint

MaicolBen:
This aims to improve the doc work in order to reuse the request/response payloads.

Check https://apiblueprint.org/documentation/mson/specification.html for more info

You can check https://railsapibase.docs.apiary.io/# for the preview

In the future, [Api Blueprint expects to make headers reusable](https://github.com/apiaryio/api-blueprint-rfcs/pull/3), so we don't have to repeat the auth headers every time.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -10,22 +10,8 @@ Rails Api Base is a boilerplate project for JSON RESTful APIs. It follows the co
 ### Sign Up [POST]
 
 + Request (application/json)
-    + Body
-
-            {
-              "user":
-                {
-                  "username": "test",
-                  "email": "test@gmail.com",
-                  "password": "password",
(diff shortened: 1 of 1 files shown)


apiary.apib
```diff
@@ -265,10 +178,11 @@ Rails Api Base is a boilerplate project for JSON RESTful APIs. It follows the co
 
 ## Reset password [/api/v1/users/password]
 
+More information of how reset password works: https://github.com/lynndylanhurley/devise_token_auth/wiki/Reset-Password-Flow
+
 ### Reset passowrd [POST]
```
santiagovidal: Typo here.

MaicolBen:
Yes, you can do `array[User]`

PR #107, opened 2017-12-21
Override devise render error to be consistent with the API

### LINKED ISSUE #101: Returns error key instead of errors when there is only one error
Nowadays, when there is any kind of error, the base returns a json response with:
- `error` key in case there is only one error. Example: https://github.com/rootstrap/rails_api_base/blob/master/app/controllers/api/v1/api_controller.rb#L27
- `errors` key in case there are more than one error: Example: https://github.com/rootstrap/rails_api_base/blob/master/app/controllers/api/v1/api_controller.rb#L37

However, we are using `devise_token_auth_gem` gem and they return a json response with the key `errors` doesn't matter if there is only one error or more than one. Example: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/sessions_controller.rb#L131

The proposed solution is override `devise_token_auth` render error methods and adjust them to our current standard.

GuilleLeopold:
Resolves #101

This PR handles this issue #101 overriding the render_error method from devise (added by @MaicolBen in this PR https://github.com/lynndylanhurley/devise_token_auth/pull/989 ) to be more consistent with the way the API render the errors or an error

### DIFF
--- a/app/controllers/api/concerns/act_as_api_request.rb
+++ b/app/controllers/api/concerns/act_as_api_request.rb
@@ -18,6 +18,13 @@ module Api
         # http://stackoverflow.com/a/12205114/2394842
         request.session_options[:skip] = true
       end
+
+      def render_error(status, message, _data = nil)
+        response = {
+          error: message
+        }
+        render json: response, status: status
+      end
     end
   end
 end
--- a/spec/requests/api/v1/sessions/create_spec.rb
+++ b/spec/requests/api/v1/sessions/create_spec.rb
@@ -58,8 +58,7 @@ describe 'POST api/v1/users/sign_in', type: :request do
 
       expect(response).to be_unauthorized
       expected_response = {
-        success: false,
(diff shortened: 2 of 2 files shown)


app/controllers/api/concerns/act_as_api_request.rb
```diff
@@ -18,6 +18,13 @@ def skip_session_storage
         # http://stackoverflow.com/a/12205114/2394842
         request.session_options[:skip] = true
       end
+
+      def render_error(status, message, _data = nil)
```
santiagovidal: Why here and not api_controller.rb?

MaicolBen: Because the devise token auth controllers don't inherit from the api controller

santiagovidal: That's very true

PR #120, opened 2018-04-18
Upgrade to Rails to 5.2

MaicolBen:
Check the whole [changelog](http://guides.rubyonrails.org/5_2_release_notes.html#upgrading-to-rails-5-2)

This PR includes:
- Upgraded rails and some other gems
- Add bootsnap that is now by default in rails (it speeds up the boot of the apps)
- Removed `tilt` from the gemfile, to update a dependency isn't needed to declare it in the gemfile, just `bundle update my_gem` does it.

TODO:
- [x] Remove success deprecation when [rspec releases new version](https://github.com/rspec/rspec-rails/pull/1951)
- [x] Wait until active admin release a new version under 1.2.x

An important change in Rails 5.x is that rails [isn't halting callbacks with false anymore](https://blog.bigbinary.com/2016/02/13/rails-5-does-not-halt-callback-chain-when-false-is-returned.html)


grilix:
What else do we need to get this merged, @MaicolBen?

MaicolBen:
@grilix rspec rails needs to bump v3.7.2, right now there are tons of warnings when you run the tests, it's already fixed in rspec-rails

grilix:
Thanks, @MaicolBen! :+1:

PR #127, opened 2018-07-02
Remove Facebook & Action Cable

MaicolBen:
This will be supported in https://github.com/rootstrap/rails_base

This should be squashed when merging with a better commit message


santiagovidal:
We should update the readme.

matiasmansilla1989:
@MaicolBen @fedeagripa @santiagovidal Should't we keep a facebook branch and maybe an Action Cable one? We should have some countermeasures in case something fails when we start using the new gem.

MaicolBen:
@matiasmansilla1989 You can go back to this PR if you need Facebook. And if there is a bug in `rails-base` it should be fixed there, and if we cannot fix it there, we can always revert this PR in the future.

matiasmansilla1989:
@MaicolBen Yes, you are right. I think most of the company know that we used to have facebook and it's obvious that you can go to previous PR.

fedeagripa:
@santiagovidal readme updated

PR #130, opened 2018-07-16
Prevent app failure because figaro required vars

fedeagripa:
When projects start with application.yml from the example it fails because figaro required vars. Adding empty strings prevent it to fail when you don't use them at the start

### DIFF
--- a/config/initializers/figaro.rb
+++ b/config/initializers/figaro.rb
@@ -3,13 +3,7 @@ variables = %w[SERVER_URL PASSWORD_RESET_URL]
 
 unless Rails.env.test?
   # Variables not used by the test environment
-  variables += %w[SECRET_KEY_BASE
-                  SENDGRID_USERNAME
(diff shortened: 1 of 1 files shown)


santiagovidal:
As we discussed, let's go with the other approach (remove the keys from `figaro.rb`)

fedeagripa:
@santiagovidal @matiasmansilla1989 fixed

config/initializers/figaro.rb
```diff
@@ -1,15 +1,3 @@
 # Put here the variables used by all the environments
 variables = %w[SERVER_URL PASSWORD_RESET_URL]
-
-unless Rails.env.test?
-  # Variables not used by the test environment
-  variables += %w[SECRET_KEY_BASE
```
santiagovidal: We should keep `SECRET_KEY_BASE` because it's not related to any external service.

Maybe the best approach is to keep the structure of the code and just delete the variables below.

Issue #134, opened 2018-08-08
ActionViewTemplateError when record is not found in a view


GuilleLeopold:
When: 
When trying to call a helper method inside a view and the helper hasn't been called before

What:
If the record is not found, it raises an error that it's catched by the `render_error`  method in the ApiController instead of the `render_record_not_found` method. 

Why:
When raising the error in the view, instead of raising an ActiveRecord:RecordNotFound error, it's raising an `ActionViewTemplateError` and that's why is not being catched by the `render_record_not_found`

For further details, ask @jdrosales17 who is the one that found the issue

santiagovidal:
Fixed by https://github.com/rootstrap/rails_api_base/pull/135

PR #135, opened 2018-08-09
Return 404 when RecordNotFound happens in a view

santiagovidal:
Solves https://github.com/rootstrap/rails_api_base/issues/134

The proposed solution is to use the exception's message to check if the API should return a 404.

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -28,6 +28,11 @@ module Api
       def render_error(exception)
         raise exception if Rails.env.test?
 
+        # To properly handle RecordNotFound errors in views
+        if exception.cause.is_a?(ActiveRecord::RecordNotFound)
+          return render_not_found(exception)
+        end
+
         logger.error(exception) # Report to your error managment tool here
         render json: { error: I18n.t('api.errors.server') }, status: 500 unless performed?
       end


santiagovidal:
To all the reviewers: I added a new commit that deletes an unused method in `ActAsApiRequest` concern

app/controllers/api/v1/api_controller.rb
```diff
@@ -28,6 +28,12 @@ def status
       def render_error(exception)
         raise exception if Rails.env.test?
 
+        # To properly handle RecordNotFound errors in views
+        if exception.message.starts_with?("Couldn't find")
+          render_not_found(exception)
+          return
```
letiesperon: I don't think this return is needed.

app/controllers/api/v1/api_controller.rb
```diff
@@ -28,6 +28,12 @@ def status
       def render_error(exception)
         raise exception if Rails.env.test?
 
+        # To properly handle RecordNotFound errors in views
+        if exception.message.starts_with?("Couldn't find")
```
grilix: Can't we do something like `if exception.is_a?(RecordNotFound)` instead?

santiagovidal: As far as I know, no. The exception's class is `ActionView::Template::Error`

app/controllers/api/v1/api_controller.rb
santiagovidal: Fixed 👍

app/controllers/api/v1/api_controller.rb
santiagovidal: Correction: as we discussed, we can use `cause`. Fixed.

PR #136, opened 2018-10-01
Sign out spec now checks if the token was destroyed

santiagovidal:
The idea here is to improve the sign out specs by checking if the token was deleted. I also added a test for the case when no token is received.

Credit to @leandrodzp, who did it first in this PR https://github.com/leandrodzp/target2/pull/11

### DIFF
--- a/spec/requests/api/v1/sessions/destroy_spec.rb
+++ b/spec/requests/api/v1/sessions/destroy_spec.rb
@@ -3,15 +3,24 @@ require 'rails_helper'
 describe 'DELETE api/v1/users/sign_out', type: :request do
   let(:user) { create(:user) }
 
-  it 'returns a successful response' do
-    delete destroy_user_session_path, headers: auth_headers, as: :json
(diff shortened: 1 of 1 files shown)


spec/requests/api/v1/sessions/destroy_spec.rb
```diff
@@ -3,15 +3,24 @@
...
+  context 'with a valid token' do
+    it 'returns a successful response' do
+      delete destroy_user_session_path, headers: auth_headers, as: :json
+      expect(response).to have_http_status(:success)
+    end
+
+    it 'decrements the amount of user tokens' do
+      headers = auth_headers
+      expect do
+        delete destroy_user_session_path, headers: headers, as: :json
```
gabrielbursztein2: Why not headers: auth_headers?

letiesperon: Because `auth_headers` adds the new headers to the user, so if the expect block measures the tokens changing they will change by 0

PR #137, opened 2018-10-12
fix default belongs_to presence validation

GuilleLeopold:
This PR is intended to fix an issue with the new_framework_default initializer which is supposed to have the configuration to make sure that a belongs_to association validates the presence of the record by default. 

Right now as we check it with @jdrosales17  and @JuanfraM this is not working and with this change we fix that issue

### DIFF
--- a/config/initializers/new_framework_defaults.rb
+++ b/config/initializers/new_framework_defaults.rb
@@ -9,7 +9,7 @@
 ActiveSupport.to_time_preserves_timezone = true
 
 # Require `belongs_to` associations by default. Previous versions had false.
-Rails.application.config.active_record.belongs_to_required_by_default = true
+ActiveRecord::Base.belongs_to_required_by_default = true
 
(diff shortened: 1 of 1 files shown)


matiasmansilla1989:
Good work. However it drew my attention since that seems to be a rails configuration generated automatically by Rails, do you have any link/reference talking about this?

letiesperon:
I had the same problem. It is from a gem stepping over the configuration. Another way to fix this (I think it's cleaner is to add it in `application.rb` since it's that configuration. `config.active_record.belongs_to_required_by_default = true`

GuilleLeopold:
https://github.com/rails/rails/issues/23589#issuecomment-305557647 here it's the issue @matiasmansilla1989 . I prefer to keep it in that file as it's the way Rails opted to use. Anyway it'seems to be fixed in Rails 5.1 by loading the defaults config for rails 5.1

Issue #138, opened 2018-10-17
ActiveStorage


santib:
I think it'd be a good idea to move from Carrierwave to ActiveStorage since it's the Rails way to manage files. I guess we can try it with one app before making the change here.

https://edgeguides.rubyonrails.org/active_storage_overview.html

Thoughts?

matiasmansilla1989:
I agree with this idea but taking into consideration what @santiagovidal came up in Slack, we should support base 64 as well. Apparently, it's easy to do it manually so we can go ahead with this.
Great suggestion, thanks.

PR #141, opened 2018-11-02
Use Active Storage instead of Carrierwave and Fog

santib:
This PR removes Carrierwave and Fog and configures ActiveStorage. This change was discussed https://github.com/rootstrap/rails_api_base/issues/138 and the main reason to go with ActiveStorage is that it's the "rails way" to work with attachments.

It also adds a helper method to support attaching base64 encoded images.


lib/base64_attach.rb
```diff
@@ -0,0 +1,10 @@
+module Base64Attach
```
letiesperon: Just for curiosity, where is this used? :)

santib: Nowhere yet, it will be used wherever you want to attach a base64 image, I guess this will be mostly in controllers.

santib: The idea was to simplify the attachment of base64 images, allowing to do something like this `user.image.attach(from_base64(params[:user][:image]))`

santiagovidal:
I added more reviewers so everyone is aware of this change.

lib/base64_attach.rb
```diff
@@ -0,0 +1,10 @@
...
+  extend self
+
+  def from_base64(data:, filename:)
```
grilix: After some discussion with @santib, since we could expect this module to be included into a controller, I'd like to find a better name for this method. The name `from_base64` is bit too generic, and we would probably hesitate to change these things after it's on the upstream.

I think `attachment_from_base64` could be a good fit, what do you people think?

In addition, it seems like reek doesn't like the name ending with `64`, can we remove that rule? Ruby itself uses names ending with numbers, so I don't think it's something bad.

letiesperon: I think reek doesn't like names to end in numbers so you don't do things like `user1`, `user2`, to differentiate params so I wouldn't remove that rule completely but this specific case is different. I would ignore only this case (Just my opinion)

grilix: I don't really see any problem with `user1` and `user2`, it sounds like depending on the context, could be correct names. Why can we use `first_user`, `second_user`, `user_one`, `user_two`, but not those? I think those restrictions are unnecessary.

Anyways, adding an exception for this method could be enough for now.

santib: I'm good both with the name `attachment_from_base64` and excluding the reek check for this method.

santib:
I just made the requested changes:
- method rename
- reek to ignore the method for the UncommunicativeMethodName lint

Additionally, I added to the same method:
- ability to strip headers from base64 images if they are present
- default filename
- some tests

spec/unit/base64_attach_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
```
santiagovidal: I'm more inclined to put this under a `lib` folder instead of unit. What do the others think?

santib: Done

santiagovidal:
Are we ready to merge this?

PR #142, opened 2018-11-23
Convert to rspec 3 some rspec 2 specs

gonzaM:
#### Description:

* Change to RSpec 3 syntax some specs that used RSpec 2. 
The gem [transpec](https://github.com/yujinakayama/transpec) was used to ensure that there is no missing RSpec 2 examples.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -34,12 +34,12 @@ require 'rails_helper'
 describe User do
   describe 'validations' do
     subject { build :user }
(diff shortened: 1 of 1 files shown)


santib:
Is there more RSpec 2 stuff we want to migrate? If that's the case I'd just run `transpec` https://github.com/yujinakayama/transpec . What do you think?

gonzaM:
I already run it and seems that this file is the only that needs to be fixed.

PR #144, opened 2018-12-13
Enforce the usage of braces for chaining

santib:
Seems like it's something we already decided about in the [tech guides](https://github.com/rootstrap/tech-guides/tree/master/ruby#single-line-blocks), so I'm adding the corresponding rule here.

Personally, I agree it's nicer to user braces when chaining. Example:
```ruby
expect do
  run_job
end.to change(User, :count).by(1)

# or

expect {
  run_job
}.to change(User, :count).by(1)
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -62,3 +62,6 @@ Style/FrozenStringLiteralComment:
 
 Style/ModuleFunction:
   Enabled: false
+
+Style/BlockDelimiters:
+  EnforcedStyle: braces_for_chaining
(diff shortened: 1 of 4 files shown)


santib:
@grilix I agree when the block has only one method to put it inline as you show above, my example wasn't the best.

When there are more than one line in the chained block I just prefer the braces over the `let` or the `do...end`. I guess it's just a matter of taste 🤷‍♂️

grilix:
I guess we have enough upvotes for merging it. Thank you, everyone.

PR #145, opened 2018-12-18
Add aws-sdk-s3 gem

santib:


Gemfile
```diff
@@ -5,6 +5,7 @@ ruby '2.4.1'
 gem 'rails', '~> 5.2.0'
 
 gem 'activeadmin', '~> 1.3.0'
+gem 'aws-sdk-s3', require: false
```
grilix: Adding the version will help making sure we use the correct one and avoid future issues.

Issue #139, opened 2018-10-17
DatabaseCleaner


santib:
Should we remove Database Cleaner? It seems to be unnecessary having a default Rails app.
https://anti-pattern.com/transactional-fixtures-in-rails

santib:
I think after Rails 5.1.5 DatabaseCleaner is not necessary even when using Capybara.

https://guides.rubyonrails.org/5_1_release_notes.html#system-tests
https://stackoverflow.com/questions/49246124/is-databasecleaner-still-necessary-with-rails-system-specs

PR #146, opened 2018-12-18
Remove Database Cleaner

jdrosales17:
With @santib, we removed Database Cleaner in The List project, which has Rails 5.2 and some Capybara tests, and everything worked fine (all the tests passed).

As @santib explains in https://github.com/rootstrap/rails_api_base/issues/139, from Rails 5.1.5 this gem is not needed anymore, so this PR removes the use of it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,7 +49,6 @@ group :development do
 end
 
 group :test do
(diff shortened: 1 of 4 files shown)


grilix:
We have enough approves for merging this one as well. Good job!

PR #150, opened 2019-01-14
add codeclimate

JuanfraM:
This PR includes:

* Configuration to enable test coverage in codeclimate
* Also it was added the new version of .codeclimate.yml

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,9 @@
 version: 2
 jobs:
   build:
+    environment:
+      # You must replace the CC_TEST_REPORTER_ID with the one of your project
+      CC_TEST_REPORTER_ID: 89b5ee3c59b8aeeb88ddc2f5469410dc9e0fb800ec4346a54a759bfc4eccb6b3
     docker:
       - image: circleci/ruby:2.4.1-node-browsers
       - image: circleci/postgres:9.4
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -40,3 +43,14 @@ jobs:
...
+
+      # run codeclimate!
+      - run:
+          name: Setup Code Climate test-reporter
+          command: |
+            curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+            chmod +x ./cc-test-reporter
+            ./cc-test-reporter before-build
+            ./cc-test-reporter after-build
+
```
matiasmansilla1989: remove this extra end of lines

.codeclimate.yml
```diff
@@ -7,21 +49,22 @@ engines:
...
-ratings:
-  paths:
-  - app/**
-  - lib/**
-  - "**.rb"
+    config:
+      file: '.rubocop.yml'
+    exclude_patterns:
+      - db/**/*
+exclude_patterns:
```
santib: does this exclude files for every check? If that's correct, why are we also excluding these same patterns on specific checks?

JuanfraM: I'll check it. thanks!

PR #152, opened 2019-02-11
Use active_storage_base64

santib:
Use our own gem 😄


fedeagripa:
@santib I was planning to do a pr to migrate to rails 6 and add this gem, do you mind solving the conflicts or may I ?

santib:
@fedeagripa you can do it if you want

Gemfile
```diff
@@ -4,6 +4,7 @@ ruby '2.4.1'
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.2.2'
 
+gem 'active_storage_base64'
```
Alxzu: Why not pointing v 0.1.3?

santiagovidal: We won't be uploading many new versions to it, but for consistency maybe we can specify the version here.

PR #153, opened 2019-02-26
Update the versions of all linters

santiagovidal:
This PR updates the versions of the following linters:
- `rubocop`
- `reek`
- `rails_best_practices`
- `brakeman`

Of those updates, the major change is `reek`, which we are moving from version 4 to 5. There are some important changes to `rubocop` as well.

Changes to `reek`:
- **Important** Version bumped from `4.7.2` to `5.3.1`
- New config file: `.reek.yml`
- I added a few checks to the config file that were missing, but keeping the default settings.

Changes to `rubocop`:
- Version bumped from `0.49.1` to `0.65.0`
- Enabled the `Layout/ClassStructure` (disabled by default).
- Enabled `Style/ReturnNil` (disabled by default).
- Disabled `Style/ExpandPathArguments` (enabled by default).
- **Important** Enabled the `Rails` cops (disabled by default). Of all the Rails cops, the following cops are different from the default configuration:
  - `Rails/FilePath` is disabled
  - `Rails/SaveBang` is enabled


.reek.yml
```diff
@@ -0,0 +1,128 @@
...
+  ModuleInitialize:
+    enabled: true
+    exclude: []
+  NestedIterators:
+    enabled: true
+    exclude: []
+    max_allowed_nesting: 2
+    ignore_iterators: []
+  NilCheck:
+    enabled: false
```
grilix: Any chance we can enable this one? We see a lot of nil checks :thinking:

.reek.yml
```diff
@@ -0,0 +1,128 @@
...
+    exclude:
+      - Base64Attach#attachment_from_base64
+    reject:
+      - "/^[a-z]$/"
+      - "/[0-9]$/"
+      - "/[A-Z]/"
+    accept: []
+  UncommunicativeModuleName:
+    enabled: true
+    exclude: ['V1']
```
grilix: Can we have this one as a regex? Something like, for example `/V\d/`? Not sure on how it's going to resolve the conflict with the reject list, tho.

.reek.yml
santiagovidal: It makes sense in theory, but the rule is just too aggressive. For example, in `ActAsApiRequest` it was complaining about this line: `request_content_type&.match?(/json/)`

Basically it will complain if we use safe navigation or `try`.

grilix: Okay, yeah.. I really don't like the `try`, but I think the `&.` is not that bad and we should allow it. Thanks for the clarification.

.reek.yml
santiagovidal: Fixed 👍

PR #155, opened 2019-05-10
Limit user simultaneous clients

letiesperon:
# Limit user simultaneous clients 

## Description

Users might not ever sign out, but simply delete cookies, which will keep accumulating client tokens on the `user` record. This makes the requests for the users with a big number of tokens really slow, due to all the time it takes to iterate over and decode the tokens for comparison. 

---

## 📌 Tasks

* Updates devise token auth gem (previous version of the gem had a bug unsolved where the configuration `max_clients_allowed` was not taking into consideration on sign in requests. 
* Configures the variable `max_clients_allowed` to 5. (I chose 5 because I thought it makes sense but if you want to lower it it's fine too). 
* Also fixes typo on factory name (should be plural) so the `annotate` gem annotates it. 

---


letiesperon:
@juanmanuelramallo Nice catch! I don't fully understand why tests are passing then (some check that the model is invalid when creating user with devise). How can I reproduce it? What fix can you think of?

juanmanuelramallo:
> @juanmanuelramallo Nice catch! I don't fully understand why tests are passing then (since some tests check that the model is invalid when creating user with devise (this raises an error that is caught)). How can I reproduce it? What fix can you think of?

Hmm I just found this https://github.com/rootstrap/rails_api_base/blob/master/app/controllers/api/concerns/act_as_api_request.rb#L24
Seems like we are all good. 

But in any case (to avoid clashing with that gem) I'd rename our `render_error` method, to something like `render_base_error` or `render_common_error` or something 🙈.

letiesperon:
@juanmanuelramallo I suspect part of the idea of the method is to override it this way. Check how the method renders the error: https://github.com/lynndylanhurley/devise_token_auth/blob/7f8a0301191c4a857c94a5fcb2029de04bcf0d00/app/controllers/devise_token_auth/application_controller.rb#L70
it's not how our frontends are used to handling ours. Thoughts?

juanmanuelramallo:
> @juanmanuelramallo I suspect part of the idea of the method is to override it this way. Check how the method renders the error: https://github.com/lynndylanhurley/devise_token_auth/blob/7f8a0301191c4a857c94a5fcb2029de04bcf0d00/app/controllers/devise_token_auth/application_controller.rb#L70
> it's not how our frontends are used to handling ours. Thoughts?

Excellent, just assume I didn't say anything then! 😅 
It all looks good!

letiesperon:
@juanmanuelramallo Nice catch anyway! Thank you!

PR #156, opened 2019-05-20
Several minor improvements

santiagovidal:
In this PR:
- Removed `CIRCLE_ARTIFACTS` because we are now using CodeClimate to visualize code coverage.
- Remove old exclusion from `.reek.yml`
- Check if development environment when creating sample admin user in seeds.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,1 +1,1 @@
-AdminUser.create!(email: 'admin@example.com', password: 'password')
+AdminUser.create!(email: 'admin@example.com', password: 'password') if Rails.env.development?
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -1 +1 @@
-AdminUser.create!(email: 'admin@example.com', password: 'password')
+AdminUser.create!(email: 'admin@example.com', password: 'password') if Rails.env.development?
```
letiesperon: Does it make any sense to keep on seeding in every test then?
Maybe we should change the approach we are using on our projects to make tests faster. Instead of always seeding before a test, invoking the seeds on the tests where we need them, thoughts?

santiagovidal: I'm not sure if we are seeding before every test 🤔

letiesperon: You are right, I checked and we are not. Got confused because I saw it in many projects 👍

PR #157, opened 2019-06-03
Enhancement - Migrate linters to CodeClimate

mdornel:
Since CodeClimate supports Reek plugin we are migrating `reek` check from CircleCI to CodeClimate.

- Now in `.circleci/config.yml` only includes `rails_best_practices` check which is the only linter that isn't supported by CodeClimate.

- Now `.codeclimate.yml` includes `brakeman` , `rubocop` and  `reek`.

- We still keep `code_analysis` rake task in order to run it locally.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -64,3 +64,7 @@ plugins:
       file: '.rubocop.yml'
     exclude_patterns:
       - db/**/*
+  reek:
+    enabled: true
+    config:
+      file: '.reek.yml'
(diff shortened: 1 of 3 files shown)


.codeclimate.yml
```diff
@@ -64,3 +64,7 @@ plugins:
       file: '.rubocop.yml'
     exclude_patterns:
       - db/**/*
+  reek:
```
santiagovidal: We should specify the same version we are using in the Gemfile. Same for `brakeman` and `rubocop` (it has 0.49 but it should be 0.65)

PR #158, opened 2019-06-19
update from ruby2.4.1 to ruby2.5.1

GuilleLeopold:
Needed to:
- Update oj gem
- Remove binding_of_caller as it was breaking when trying to run the server or the console https://github.com/BetterErrors/better_errors/issues/411


Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.4.1'
+ruby '2.5.1'
```
santib: why `2.5.1` and not `2.5.5`(highest patch version) or even `2.6.3` ?

santiagovidal: I would suggest updating to `2.5.5` in this PR, and we can create an issue to update to `2.6.3`. WDYT? @santib @GuilleLeopold

PR #159, opened 2019-07-15
Update Rubocop (use rubocop-rails)

santib:
1. Removed `rubocop` in favor of `rubocop-rails` (last version).
2. Required `rubocop-rails` in `.rubocop.yml` config file.
3. Excluded `db/schema.rb` from ALL rubocop linters, since it's an autogenerated file.
4. Ran `bundle exec rubocop -a .` and some errors were fixed automatically.
5. Manually fix remaining issues.

Note that I added two rules
```
Style/SymbolArray:
  EnforcedStyle: brackets

Style/WordArray:
  EnforcedStyle: brackets
```

The codebase in some places is using `[:something]`, in others `%i(something)` and in others `%i[something]` . I chose the first option just because I think they are nicer and easier, but if someone prefer other option I can revert that change.


gonzaM:
@santib I think is better to use `%i[]` and `%w[]` since the rootstrap guides prefer that.
We can discuss which one is better but first should update the guides. WDYT?

santib:
> @santib I think is better to use `%i[]` and `%w[]` since the rootstrap guides prefer that.
> We can discuss which one is better but first should update the guides. WDYT?

I agree to follow the guides 👍  I prefer the brackets style but I don't think it's something we can discuss, it's just a matter of preference. I don't think there can be real arguments in that discussion 😂  It's like the never ending semicolon vs no semicolon discussion 🤷‍♂

.rubocop.yml
```diff
@@ -49,6 +57,11 @@ Metrics/LineLength:
   # To make it possible to copy or click on URIs in the code, we allow lines
   # containing a URI to be longer than Max.
   AllowURI: true
+  Exclude:
+    - db/migrate/20161011184702_devise_create_admin_users.rb
+    - db/migrate/20161011151353_devise_create_users.rb
+    - db/migrate/20161027190856_create_delayed_jobs.rb
+    - db/migrate/20181102142200_create_active_storage_tables.active_storage.rb
```
gabrielbursztein2: Shouldn't we exclude all future migrations from this cop?

santib: I consider migrations as part of the app code and think we should care about it's quality as well. If we exclude every migration file we'd be decreasing quality there. That said, I don't have a strong opinion on any option, not sure what the rest of the people think, should we:
1. Fix those migrations
2. Exclude those migrations (like it's now)
3. Exclude every migration
?

santib:
> How this version will work with CodeClimate?
> 
> If there is a related plugin for this version on CodeClimate we should update `.codeclimate.yml` as well. Otherwise we have to run rubocop command on CircleCI.
> 
> WDYT @santiagovidal ?

Good catch, I didn't notice the rubocop version was hardcoded in `.codeclimate.yml` . Looking https://github.com/codeclimate/codeclimate-rubocop/branches/all?page=1&query=channel%2Frubocop&utf8=%E2%9C%93 I guess we can set it to use 0.71 that it's the latest available. What do you think?

mdornel:
> @mdornel Good catch, I didn't notice the rubocop version was hardcoded in `.codeclimate.yml` . Looking https://github.com/codeclimate/codeclimate-rubocop/branches/all?page=1&query=channel%2Frubocop&utf8=%E2%9C%93 I guess we can set it to use 0.71 that it's the latest available. What do you think?

My only concern is that we will be running different versions on rubocop, one version locally and other version in the integration tool. The desirable situation would be run the same rubocop version on both.

santib:
> My only concern is that we will be running different versions on rubocop, one version locally and other version in the integration tool. The desirable situation would be run the same rubocop version on both.

Yes sure, we can either put this PR on hold until codeclimate has rubocop version 0.72 available or use 0.71 version of rubocop in this PR (that is the newest version they have available), whatever you prefer.

mdornel:
> Yes sure, we can either put this PR on hold until codeclimate has rubocop version 0.72 available or use 0.71 version of rubocop in this PR (that is the newest version they have available), whatever you prefer.

Let's go with `0.71` version 👍

santiagovidal:
I agree, let's go with 0.71

santiagovidal:
I think we can merge this, right?

PR #164, opened 2019-09-12
Remove status route from users scope

glm4:
### Changed status endpoint.

* Removed `api/v1/status` from users scope in routes.


santiagovidal:
I think we can merge this now.

Issue #160, opened 2019-07-26
Reek and brakeman are not running in CodeClimate


santiagovidal:
Reek is currently not running in CodeClimate because the channel that we specified (`reek-5-3-1`) is not available.

Apparently the only channel available is `stable`, which means that we don't have control over what version of reek to run in CodeClimate.

For this reason, I believe it would be best to remove it from CodeClimate and run it only in CircleCI, same as Rails Best Practices

santiagovidal:
`brakeman` has the same problem

santiagovidal:
Fixed by https://github.com/rootstrap/rails_api_base/pull/167

PR #166, opened 2019-09-27
Remove channel specification for Brakeman, Reek and Rubocop CC plugins

mdornel:
Remove channel specification for the following CodeClimate plugins:
- Brakeman
- Reek
- Rubocop

Based on this issue https://github.com/rootstrap/rails_api_base/issues/160 we are removing channels specifications in favor of using the `stable` version of each plugins.

According to this [doc](https://docs.codeclimate.com/docs/engine-channels), leaving the channel field unspecified, an analysis runs the stable channel of a plugin.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -43,7 +43,6 @@ checks:
 plugins:
   brakeman:
     enabled: true
-    channel: brakeman-4-45
   bundler-audit:
     enabled: true
   csslint:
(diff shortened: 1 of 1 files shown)


mdornel:
We are closing this PR because the `stable` version Reek plugin in CodeClimate is not working as expected. Is showing us issues that are not being reported for Reek 5.3.1 (which is included in our Gemfile)

PR #167, opened 2019-09-30
Move back code_analysis task to Circle CI

mdornel:
- Remove code analysis tools from CodeClimate since we are having issues with the channels for `Rubocop`, `Reek` and `Brakeman` plugins. At the moment there is a mismatch between the plugins version in CC and the versions in the Api Base Gemfile.

- run `RAILS_ENV=test bundle exec rake code_analysis` as a CircleCI step.

Note: This is a temporary workaround until we find a better way or when we have more plugins versions available in channels supported by CodeCimate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,10 +34,10 @@ jobs:
           name: run tests
           command: RAILS_ENV=test bundle exec rspec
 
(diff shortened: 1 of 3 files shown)


santiagovidal:
I think we can merge this now.

PR #168, opened 2019-10-07
Exclude specs when measuring test coverage

mariosouto:
## Description

Test coverage should reveal which lines of code are we missing to test, it shouldn't measure that on the tests itself.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -50,4 +50,34 @@ describe User do
       expect(full_name).to eq(user.username)
     end
   end
+
+  context 'when user has first_name' do
+    let!(:user) { create(:user, first_name: 'John', last_name: 'Doe') }
+
+    it 'returns the correct name' do
+      expect(user.full_name).to eq('John Doe')
+    end
+  end
+
+  describe '.from_social_provider' do
+    context 'when user does not exists' do
+      let(:params) { attributes_for(:user) }
+
+      it 'creates the user' do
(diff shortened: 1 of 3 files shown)


santib:
@mariosouto Reading the documentation of simplecov, shouldn't we be using `SimpleCov.start 'rails'` instead? It automatically exclude test files (but it also count other files that are now being ignored)

mariosouto:
@santib Agree, nice catch. I'll change to that

santiagovidal:
Looks like you have to make some changes, because the total coverage went down. There's probably some folders that were ignored in the past, that aren't in the new configuration.

jpascual1994:
@santib What about the total coverage? Is it correct?

santib:
> @santib What about the total coverage? Is it correct?

I don't know, when I try to access that link I get a 500. I'm ok either fixing the coverage here or in a separate PR.

juanmanuelramallo:
@mariosouto @andresg4 Any updates on this?

mariosouto:
@juanmanuelramallo i have it on my pipeline, i'll resolve this by next week i hope

mariosouto:
Sorry it took me so long to update this, i didn't use the following config

```rb
require 'simplecov'
SimpleCov.start 'rails'
```

Mainly because it doesn't count some folders that we typically use in our projects, for example `services/`, `queries/`, etc.

I added some tests as well.

Let me know your thoughts

Issue #172, opened 2019-11-11
Update dependencies


santiagovidal:
I'm creating this issue so we do a general update on our dependencies. It's been a while since we updated them.

Some examples are:
- `draper`
- `jbuilder`
- `pg` (this one in particular)
- `activeadmin`

PR #173, opened 2019-11-13
dependencies update

t-romani:
issue #172 

run `bundle update` on terminal to update every gem
run into a couple issues:
- first when running rails s, it would fail with
```
Expected to find a manifest file in `app/assets/config/manifest.js` (Sprockets::Railtie::ManifestNeededError)
But did not, please create this file and use it to link any assets that need
to be rendered by your app:

Example:
  //= link_tree ../images
  //= link_directory ../javascripts .js
  //= link_directory ../stylesheets .css
and restart your server
```
so following [this](https://github.com/rails/sprockets-rails/issues/444) I included the file and it worked.

- second, brakeman would report
```
Confidence: High
Category: Path Traversal
Check: SprocketsPathTraversal
Message: sprockets 4.0.0 has a path traversal vulnerability (CVE-2018-3760). Upgrade to sprockets 4.0.0.beta8 or newer
```
so followed up [this](https://www.redmine.org/issues/32223) thread to solve it, downgrading the gem.
Also [this](https://blog.heroku.com/rails-asset-pipeline-vulnerability) can be informative.

- Updated `Faker::Number` usage on user factory.


santiagovidal:
Some gems like `activeadmin` and `annotate` can also be updated, but we have to change the `Gemfile`.

`annotate`'s latest version is 3.0.3
`activeadmin`'s latest version is 2.4.0

There may be more like them.

app/assets/config/manifest.js
```diff
@@ -0,0 +1 @@
+{}
```
santib: shouldn't this be
```
//= link_tree ../images
//= link_directory ../javascripts .js
//= link_directory ../stylesheets .css
```
?

config/application.rb
```diff
@@ -10,7 +10,6 @@
 require 'action_view/railtie'
 require 'action_cable/engine'
 require 'active_storage/engine'
-require 'sprockets/railtie'
```
santib: Why are we doing this change?

t-romani: Because that would import sprockets 4.0.0, which has vulnerabilities .
You can read more on links from second item in description.

app/assets/config/manifest.js
t-romani: That could be, I don't really understand the motives for this file. So I just followed a comment readable on the [link](https://github.com/rails/sprockets-rails/issues/444) on first description item, which created it that way. 
I'll do this change.

config/application.rb
santib: How can `sprockets 4.0.0` be imported if you put `gem 'sprockets', '~> 3.7.2'` in the `Gemfile`?

t-romani: 🤷‍♀ I don't know about that. It was just not taking `Gemfile` version.

santib: Maybe you were running `brakeman` without prefixing it with `bundle exec`? https://thoughtbot.com/blog/but-i-dont-want-to-bundle-exec

Gemfile
```diff
@@ -5,23 +5,24 @@ ruby '2.6.3'
...
-gem 'oj', '~> 3.7', '>= 3.7.12'
-gem 'pg', '~> 0.18.2'
-gem 'pry-rails', '~> 0.3.6'
-gem 'puma', '~> 3.0'
-gem 'rack-cors', '~> 0.4.0'
+gem 'oj', '~> 3.9', '>= 3.9.2'
+gem 'pg', '~> 1.1', '>= 1.1.4'
+gem 'pry-rails', '~> 0.3.9'
+gem 'puma', '~> 4.3'
+gem 'rack-cors', '~> 1.0', '>= 1.0.4'
```
santib: image

For some reason version 1.0.4 was yanked, we'll need to upgrade it

t-romani: Yep, done

santib: https://nvd.nist.gov/vuln/detail/CVE-2019-18978

config/application.rb
t-romani: I was running `code_analysis` task:
`task :code_analysis do
  sh 'bundle exec brakeman . -z -q'
...
end
`

santib: I tried it myself (putting back this line) and didn't get any warning message. Let me know if you want to pair on this one

t-romani: 🤔 that is certainly weird, who knows what did I do 😆 
Sure! Let's do it

Issue #171, opened 2019-11-11
Remove `rememberable` from our Devise modules


santiagovidal:
We should remove this module because we are not using it.

`rememberable` is only applicable for cookie based authentication.

PR #175, opened 2019-11-18
Issue fix: Rememberable devise module removed

nicofh:
Devise module and column `Rememberable` removed from `user` model and table.

Issue: https://github.com/rootstrap/rails_api_base/issues/171


app/models/admin_user.rb
```diff
@@ -26,5 +26,5 @@ class AdminUser < ApplicationRecord
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable,
-         :recoverable, :rememberable, :trackable, :validatable
```
santiagovidal: We think we can keep it for activeadmin, because in this case we are using cookies.

PR #178, opened 2020-01-02
Use loose ruby versioning

santib:
Idea from @juanmanuelramallo 

https://andycroll.com/ruby/use-loose-ruby-versioning-in-your-gemfile/


fedeagripa:
Nice 👏

PR #181, opened 2020-01-20
Require rails_helper from '.rspec'

martinjaimem:
Add --require rails_helper to a `.rspec` file in the project root. This has the effect that all the specs are automatically run with rails_helper pre-required.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,2 @@
+--require rails_helper
+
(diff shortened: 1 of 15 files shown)


.rspec
```diff
@@ -0,0 +1 @@
+--require rails_helper
```
santiagovidal: Add a new line here.

Issue #165, opened 2019-09-26
Feature idea: forced update


santiagovidal:
I'm creating this issue to keep track of an idea that has been going around for some time: forced updates for mobile apps.

PR #182, opened 2020-01-29
[ISSUE-165] add force update endpoint

fedeagripa:
## Issue
https://github.com/rootstrap/rails_api_base/issues/165

## Notes
- Add setting table with a key value structure to add needed configuration and further needed for other configurations
- Add initial min version needed as 0.0
- In case you need to update your version you can modify the db value and thats all

### DIFF
--- a/app/controllers/api/v1/settings_controller.rb
+++ b/app/controllers/api/v1/settings_controller.rb
@@ -1,0 +1,20 @@
+module Api
+  module V1
+    class SettingsController < Api::V1::ApiController
+      skip_before_action :authenticate_user!
+
+      def must_update
+        current_version = Gem::Version.new(params[:device_version])
+        min_version = version ? Gem::Version.new(version) : Gem::Version.new('0.0.0')
+        must_update = min_version > current_version
+        render json: { 'must_update': must_update }
+      end
+
+      private
+
+      def version
+        Setting.find_by(key: 'min_version').try(:value)
+      end
+    end
+  end
+end
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -1,0 +1,12 @@
+# == Schema Information
+#
+# Table name: settings
+#
+#  id    :integer          not null, primary key
+#  key   :string
+#  value :string
+#
+
+class Setting < ApplicationRecord
+  validates :key, uniqueness: true, presence: true
+end
--- a/db/migrate/20200128183708_add_must_update_fields.rb
+++ b/db/migrate/20200128183708_add_must_update_fields.rb
@@ -1,0 +1,8 @@
+class AddMustUpdateFields < ActiveRecord::Migration[6.0]
+  def change
+    create_table :settings do |t|
+      t.string :key, null: false
+      t.string :value
+    end
+  end
+end
--- a/spec/factories/settings.rb
+++ b/spec/factories/settings.rb
@@ -1,0 +1,20 @@
+# == Schema Information
+#
+# Table name: settings
+#
+#  id    :integer          not null, primary key
+#  key   :string
+#  value :string
+#
+
+FactoryBot.define do
+  factory :setting do
+    key   { 'key' }
+    value { 'value' }
+
+    factory :setting_version do
+      key   { 'min_version' }
+      value { '0.0' }
+    end
+  end
+end
--- a/spec/requests/api/v1/settings_spec.rb
+++ b/spec/requests/api/v1/settings_spec.rb
@@ -1,0 +1,40 @@
+require 'rails_helper'
+
+describe 'GET api/v1/settings/must_update', type: :request do
+  let(:params) do
+    {
+      device_version: '1.0'
+    }
+  end
+  subject { get must_update_api_v1_settings_path, params: params, as: :json }
+
+  context 'with correct settings' do
+    let!(:setting_version) { create(:setting_version) }
+
+    it 'return success' do
+      subject
+      expect(response).to be_successful
+    end
+
+    it 'returns no need to update' do
+      subject
+      expect(json['must_update']).to be(false)
+    end
+  end
+
+  context 'with incorrect settings' do
+    let!(:setting_version) { create(:setting_version, value: '2.0') }
+
+    it 'returns it needs to update' do
+      subject
+      expect(json['must_update']).to be(true)
+    end
+  end
+
+  context 'without setting min version record' do
+    it 'returns no need to update' do
+      subject
+      expect(json['must_update']).to be(false)
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,6 +16,9 @@ Rails.application.routes.draw do
           get :profile
         end
       end
+      resources :settings, only: [] do
+        get :must_update, on: :collection
+      end
     end
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_11_02_142200) do
+ActiveRecord::Schema.define(version: 2020_01_28_183708) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -66,6 +66,11 @@ ActiveRecord::Schema.define(version: 2018_11_02_142200) do
     t.datetime "created_at"
     t.datetime "updated_at"
     t.index ["priority", "run_at"], name: "delayed_jobs_priority"
+  end
+
+  create_table "settings", force: :cascade do |t|
+    t.string "key", null: false
+    t.string "value"
   end
 
   create_table "users", id: :serial, force: :cascade do |t|
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,1 +1,2 @@
 AdminUser.create!(email: 'admin@example.com', password: 'password') if Rails.env.development?
+Setting.create_or_find_by(key: 'min_version', value: '0.0')


app/controllers/api/v1/settings_controller.rb
```diff
@@ -0,0 +1,14 @@
+module Api
+  module V1
+    class SettingsController < Api::V1::ApiController
+      skip_before_action :authenticate_user!
+
+      def need_update
+        current_version = params[:device_version].to_f
+        min_version = Setting.find_by(key: 'min_version').value.to_f
```
jpascual1994: What happens if min_version is `0.0.1`? Does `to_f` work with it?
Maybe you can use Gem::Version.new(version)

app/controllers/api/v1/settings_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+        @must_update = min_version > current_version
```
santiagovidal: This doesn't have to be an ivar. You are only using it here.

app/models/setting.rb
```diff
@@ -0,0 +1,13 @@
...
+# Table name: settings
+#
+#  id    :integer          not null, primary key
+#  key   :string
+#  value :string
+#
+
+class Setting < ActiveRecord::Base
+  validates :key, uniqueness: true
+  validates :key, presence: true
```
santiagovidal: This can be done in the same line

spec/requests/api/v1/settings_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  # Clean settings in case you have ran seeds
+  before { Setting.destroy_all }
+  after { Setting.destroy_all }
+
+  context 'with correct settings' do
+    let!(:setting_version) { create :setting_version }
+
+    it 'return success' do
+      get need_update_api_v1_settings_path, params: params, as: :json
```
santiagovidal: You can use a variable for this to avoid repeating it

santiagovidal:
Perhaps we can implement this with an Env variable instead of a Settings table. Some projects might benefit from having the table but others won't.

Also, make sure to add as many reviewers as possible.

fedeagripa:
@santiagovidal regarding env implementation, I prefer this cause it allows to update version to non-technical people and dynamically , with env var you are forcing to update heroku and maybe restarting the dyno

app/controllers/api/v1/settings_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Api
+  module V1
+    class SettingsController < Api::V1::ApiController
+      skip_before_action :authenticate_user!
+
+      def need_update
+        current_version = Gem::Version.new(params[:device_version])
+        min_version = Gem::Version.new(version) if version
```
martinjaimem: What happens if `version` is nil?

spec/requests/api/v1/settings_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(json['must_update']).not_to be
+    end
+  end
+
+  context 'with incorrect settings' do
+    let!(:setting_version) { create :setting_version, value: '2.0' }
+
+    it 'returns it needs to update' do
+      subject
+      expect(json['must_update']).to be
```
leandrodzp: Isn't `false` necessary here?

spec/requests/api/v1/settings_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    }
+  end
+  subject { get need_update_api_v1_settings_path, params: params, as: :json }
+
+  # Clean settings in case you have ran seeds
+  before { Setting.destroy_all }
+  after { Setting.destroy_all }
+
+  context 'with correct settings' do
+    let!(:setting_version) { create :setting_version }
```
leandrodzp: Use parenthesis here `create(:setting_version)`

app/controllers/api/v1/settings_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+        must_update = min_version > current_version
+        render json: { 'must_update': must_update }
```
leandrodzp: I would use or `need_update` or `must_update` everywhere, so we use the same name at the endpoint and in the response, what do you think?

app/models/setting.rb
```diff
@@ -0,0 +1,12 @@
...
+#
+# Table name: settings
+#
+#  id    :integer          not null, primary key
+#  key   :string
+#  value :string
+#
+
+class Setting < ActiveRecord::Base
+  validates :key, uniqueness: true, presence: :true
```
leandrodzp: The `:` is wrong after `presence:` 🤔

spec/factories/settings.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+#  id    :integer          not null, primary key
+#  key   :string
+#  value :string
+#
+
+FactoryBot.define do
+  factory :setting do
+    key   { 'key' }
+    value { 'value' }
```
leandrodzp: I would use Faker here, what do you think?

app/controllers/api/v1/settings_controller.rb
fedeagripa: Fixed, thanks

spec/requests/api/v1/settings_spec.rb
fedeagripa: that matcher behaves like `be_truthy` one, but in this case due to method logic we should expect always a bool, will update thanks

spec/factories/settings.rb
fedeagripa: I won't because there is no specific one to match, also which would be the benefit of using it?

app/controllers/api/v1/settings_controller.rb
fedeagripa: I though it as the endpoint (need_update) is the question and the response (must_update)

spec/requests/api/v1/settings_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  context 'with incorrect settings' do
+    let!(:setting_version) { create(:setting_version, value: '2.0') }
+
+    it 'returns it needs to update' do
+      subject
+      expect(json['must_update']).to be(true)
+    end
+  end
+
+  context 'without setting version env var' do
```
santiagovidal: It's not an Env Var

fedeagripa: ups 😅

spec/factories/settings.rb
leandrodzp: Anyone now, but will prevent to change it in a future if we need to add extra settings and validations between them.

db/migrate/20200128183708_add_must_update_fields.rb
```diff
@@ -0,0 +1,8 @@
+class AddMustUpdateFields < ActiveRecord::Migration[6.0]
+  def change
+    create_table :settings do |t|
+      t.string :key
+      t.string :value
```
leandrodzp: I would add a non-empty validation (and their tests), what do you think?

leandrodzp: Also, I would add an index by key :).

spec/requests/api/v1/settings_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  # Clean settings in case you have ran seeds
+  before { Setting.destroy_all }
+  after { Setting.destroy_all }
+
+  context 'with correct settings' do
+    let!(:setting_version) { create(:setting_version) }
+
+    it 'return success' do
+      subject
+      expect(response).to have_http_status(:success)
```
JuanfraM: `be_successful` ?

fedeagripa:
Remember this is a base project and is not intended to add features just because we may need them. We add them as we detect we need them and we try to be as concrete as we can to don't generate boilerplate to all possible projects.

fedeagripa:
Following @sebastiancaraballo admin proposal, I don't know even if we should keep having active-admin by default and supported to all features added, given we switched to open source.
My concern is we would end up having lot of admin configuration as we add models. Also what if a project want to use other implementation of admins? we are forcing them to delete more code.
Thoughts?

spec/factories/settings.rb
fedeagripa: You should never use the default factory (only for simple model restrictions). If you are using this on a feature test is cause you want to test a specific key, so I don't find any reason to make this variable with Faker

db/migrate/20200128183708_add_must_update_fields.rb
fedeagripa: I don't agree on the index, given amount of records it will make things slower due to generating and maintaining index

app/models/setting.rb
```diff
@@ -0,0 +1,12 @@
+# == Schema Information
+#
+# Table name: settings
+#
+#  id    :integer          not null, primary key
+#  key   :string
+#  value :string
+#
+
+class Setting < ApplicationRecord
```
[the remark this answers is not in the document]
juanmanuelramallo: After looking at the whole PR it seems that an instance of setting might be anything. 

Got two comments: 
1. If we are only tackling the force update whenever the version is outdated, why not just introduce a version object for the app (like we do in gems) ?

If 1 is a no-go and you prefer current approach:
2. Why not use a single row of settings with a hstore or jsonb column containing all the settings the app might have?

juanmanuelramallo:
> @santiagovidal regarding env implementation, I prefer this cause it allows to update version to non-technical people and dynamically , with env var you are forcing to update heroku and maybe restarting the dyno

Why would we allow a non-tech person to update the version of the API? Isn't the version associated with production code changes?

app/models/setting.rb
```diff
@@ -0,0 +1,12 @@
+# == Schema Information
+#
+# Table name: settings
+#
+#  id    :integer          not null, primary key
+#  key   :string
```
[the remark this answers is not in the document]
fedeagripa: this is a key-value pair table that may be reusable for other settings, adding a table for only one record (min_version) was a big no, so this is the middle solution I think may be acceptable enough

app/models/setting.rb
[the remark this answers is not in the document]
fedeagripa: Good point, you made me remember I need to update this to split min_version into:
- min_ios_version
- min_android_version
and so you can later add your min_whatever_version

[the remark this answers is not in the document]
fedeagripa: I don't fully get the idea of hstore or jsonb, but will slack you later about this

app/controllers/api/v1/settings_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Api
+  module V1
+    class SettingsController < Api::V1::ApiController
+      skip_before_action :authenticate_user!
+
+      def must_update
```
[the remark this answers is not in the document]
fedeagripa: This will be consumed only by devices, if they don't want to use this just don't use it. Is not a must check so I don't agree on moving it inside application controller

app/models/setting.rb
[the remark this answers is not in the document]
juanmanuelramallo: And why is that we want this to be a table in the database?

app/controllers/api/v1/settings_controller.rb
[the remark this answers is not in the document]
juanmanuelramallo: What do you mean by devices? Aren't all clients of the api devices?

fedeagripa:
> > @santiagovidal regarding env implementation, I prefer this cause it allows to update version to non-technical people and dynamically , with env var you are forcing to update heroku and maybe restarting the dyno
> 
> Why would we allow a non-tech person to update the version of the API? Isn't the version associated with production code changes?

There were cases were clients wanted to release newer versions of mobile version that forced users to update, without changing API. I know usually a force update comes along with backend changes, but in some cases device changes take longer and so they can finish setting this up by themselves (modifying the value of min versions through an admin for example) when device changes are ready and released

app/controllers/api/v1/settings_controller.rb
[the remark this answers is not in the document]
fedeagripa: Android & IOS that consume the API
This is not into scope for web clients that consumes API

app/models/setting.rb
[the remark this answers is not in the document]
fedeagripa: The reason for this is to give a dynamically way update min_versions, if you can think of another way to add this with existing base code it would be great

db/migrate/20200128183708_add_must_update_fields.rb
leandrodzp: I don't agree. The index will be generated only during the insert and I don't think configuration keys will be added frequently but they should be asked often.

fedeagripa: We are talking about performance and indexing in a 2-3 rows table. If you still think is an improvement feel free to benchmark and see there will be no significant benefit if it even is

app/models/setting.rb
```diff
@@ -0,0 +1,12 @@
...
+#  value :string
```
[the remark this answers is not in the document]
fedeagripa: following top comments, the intent of being generic is to reuse if other settings are needed.
We can change to `verstion_key` & `verison_value` due to current scope, but I would prefer not to

PR #184, opened 2020-02-05
Improve ActiveJob defaults

santib:
Improve ActiveJob defaults for each environment.
Encourages usage of ActiveJob instead of using directly the DelayedJob API


config/environments/production.rb
```diff
@@ -54,8 +54,9 @@
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
 
+
```
santib:
```suggestion
```

PR #188, opened 2020-02-27
add i18n-tasks health

fedeagripa:
## WHAT
Add https://github.com/glebm/i18n-tasks gem

## WHY
This allow us to check translations health

## USAGE
Check health `i18n-tasks health`
Add missing keys `i18n-tasks add-missing`
Normalize data `i18n-tasks normalize`

## FUTURE WORK
Discuss if we want to add this as:
- automated check
- required check
- auto fixable task


fedeagripa:
@santiagovidal can you rebuild CI, don't have permissions

spec/i18n_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+                    "Please run `i18n-tasks normalize' to fix"
+    expect(non_normalized).to be_empty, error_message
+  end
+
+  it 'does not have inconsistent interpolations' do
+    error_message = "#{inconsistent_interpolations.leaves.count} i18n keys have inconsistent interpolations.\n" \
+                    "Run `i18n-tasks check-consistent-interpolations' to show them"
+    expect(inconsistent_interpolations).to be_empty, error_message
+  end
+end
```
juanmanuelramallo: Given this is not testing behavior from our production code and instead it's just about running some tasks from a vendor gem, what do you think if we have these checks in our travis CI as a new step instead of this spec?

```yml
     - run:
          name: I18n Health
          command: bundle exec i18n-tasks health
```

fedeagripa: I completely agree on this, but would love to hear at least one more opinion on this. Then I can change this in next pr 😉

.circleci/config.yml
```diff
@@ -7,6 +7,9 @@ jobs:
     docker:
       - image: circleci/ruby:2.6.3-node-browsers
       - image: circleci/postgres:9.4
+        environment:
```
santib: I think this should be indented with two more spaces (same for the two lines below)

spec/i18n_spec.rb
juanmanuelramallo: @jpascual1994 @santib Any thoughts?

jpascual1994: I totally agree 😄

.circleci/config.yml
fedeagripa: 🤔 I'm seeing this is already on master without extra spaces, so wdyt?

spec/i18n_spec.rb
jdrosales17: Agree with @juanmanuelramallo on this! We should add the health command to the CI checks and avoid this test.

jdrosales17: Please change that on this PR :)

.circleci/config.yml
santib: I was saying because of this
image

If there are other places that are incorrectly indented we could fix them too

spec/i18n_spec.rb
fedeagripa: @jdrosales17 @Juanma1991 done
Next time be dyslexia inclusive and send a copy & paste 😅

.circleci/config.yml
fedeagripa: @santib `environment` was for `docker` block so I moved it back, moving it was generating an error in the schema check

PR #195, opened 2020-03-30
Add Docker support

juan-apa:
The usage of Docker in any project is supposed to be **optional**
This PR adds the Dockerfile and docker-compose.yml files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+FROM ruby:2.6.3
+
+LABEL Name=rails_api_base Version=1.0
+
+RUN apt-get update -qq && apt-get install -y nodejs postgresql-client libxml2
+
+# Install heroku cli
+RUN curl https://cli-assets.heroku.com/install.sh | sh
+
+RUN mkdir /gems
+ENV BUNDLE_PATH "/gems"
+
+RUN mkdir /app
+WORKDIR /app
+
+# Install Node.js with npm
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt -y install nodejs
+RUN apt -y  install gcc g++ make cmake
+
+# Install yarn
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt update && apt install yarn
+
+RUN apt install nano
+
+# Install rails globally
+RUN gem install rails -v 6.0.0
+
+WORKDIR /app
+
+ADD Gemfile Gemfile.lock ./
+RUN bundle install
+RUN yarn install --check-files
+
+
+ADD . /app/
+EXPOSE 3000
+CMD rm -f tmp/pids/server.pid &&  bundle exec rails s -b 0.0.0.0 -p 3000
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@ FROM ruby:2.6.3
 
 LABEL Name=rails_api_base Version=1.0
 
+ENV PATH=/app/bin:$PATH
+
 RUN apt-get update -qq && apt-get install -y nodejs postgresql-client libxml2
 
 # Install heroku cli
@@ -16,25 +18,34 @@ WORKDIR /app
 # Install Node.js with npm
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
 RUN apt -y install nodejs
-RUN apt -y  install gcc g++ make cmake
+RUN apt -y install gcc g++ make cmake
 
 # Install yarn
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
 RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
 RUN apt update && apt install yarn
 
-RUN apt install nano
+# Install bundler
+RUN gem install bundler:2.0.2
 
-# Install rails globally
-RUN gem install rails -v 6.0.0
+# Install nano text editor
+RUN apt install nano
 
 WORKDIR /app
 
+# Add application binaries
+ADD ./bin /app/bin
+
+# Install gems
 ADD Gemfile Gemfile.lock ./
-RUN bundle install
+RUN ./bin/bundle install
+
+# Install node modules
+ADD package.json yarn.lock ./
 RUN yarn install --check-files
 
-
 ADD . /app/
+
 EXPOSE 3000
-CMD rm -f tmp/pids/server.pid &&  bundle exec rails s -b 0.0.0.0 -p 3000
+
+ENTRYPOINT bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -b 0.0.0.0 -p 3000"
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+WORKDIR /app
+
+ADD Gemfile Gemfile.lock ./
+RUN bundle install
+RUN yarn install --check-files
+
+
+ADD . /app/
+EXPOSE 3000
+CMD rm -f tmp/pids/server.pid &&  bundle exec rails s -b 0.0.0.0 -p 3000
```
Alxzu: Isn't this redundant giving the fact that we are using this command on de compose?

Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM ruby:2.6.3
```
juanmanuelramallo: Is there a way we can use whatever version we have in the `.ruby-version` file here? Just asking around

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN mkdir /gems
+ENV BUNDLE_PATH "/gems"
+
+RUN mkdir /app
+WORKDIR /app
+
+# Install Node.js with npm
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt -y install nodejs
+RUN apt -y  install gcc g++ make cmake
```
juanmanuelramallo:
```suggestion
RUN apt -y install gcc g++ make cmake
```

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+# Install yarn
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt update && apt install yarn
+
+RUN apt install nano
```
juanmanuelramallo: <3

Alxzu:
What do you think about running common tasks when starting the container? such as migrations or db creation?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+# Install rails globally
+RUN gem install rails -v 6.0.0
```
juanmanuelramallo: Do we need this? 

I mean we already have a project with a Gemfile, wouldn't it be enough just to call `bundle install` once inside the project folder?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+WORKDIR /app
+
+ADD Gemfile Gemfile.lock ./
+RUN bundle install
+RUN yarn install --check-files
+
+
```
juanmanuelramallo: Do we need this empty line?

Dockerfile
juan-apa: This is in case the image is intended to be used standalone, without docker-compose. However now that you mention it, this won't be used without a database. I'll remove it

Dockerfile
juan-apa: The problem is, that if rails is not installed globally, we need to write bundle exec everytime we want to execute a rails command e.g.: `bundle exec rails db:create`. If rails is installed globally, this command turns into the usual one: `rails db:create`. I think this removes verboseness when using rails commands. What do you think?

Dockerfile
juan-apa: probably not.

Dockerfile
juan-apa: mhhhmmm. To be honest, I don't know, I'll look into it!

juan-apa:
Also, what do you guys think about installing heroku-cli in this image? is it necessary? Usually this is the image I use for developing inside a container. However, if this image is intended to be used in production, it would be unnecessary.

Dockerfile
Alxzu: Don't we have the bin/rails command? I got the idea this command doesn't need the `bundle exec` 🤔

juanmanuelramallo: I'd rather always use `bundle exec` then :D

Alxzu: I'm not sure what would be the difference against using `bin/rails` vs `bundle exec` they seem pretty much the same, so whatever option would be fine for me

juan-apa: The bin/rails executable is present, but you would have to cd into the bin directory. If that solution is okay with you, we can skip installing rails gem globally.

Dockerfile
sebalopez: I would not remove it but replace CMD with ENTRYPOINT. 
Also you might want to provision the database separately and still run this image standalone

juan-apa: What do you mean by provisioning the database sepparately? This image only contains the rails app no database is included here. The docker compose file is the one which creates a service with the oficial postgres database. Are you suggesting that another dockerfile should be created with the postgres database?

Dockerfile
juan-apa: @juanmanuelramallo I did not find a way of doing this. If the ruby version is changed later, it will have to be changed here also. Is there a way to setup a reminder of sorts in github for this?

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: '3.0'
+
+services:
+  web:
+    image: rails_api_base
+    build: .
+    entrypoint: >
+      bash -c "
+      rm -f tmp/pids/server.pid
+      && bundle exec rails s -b 0.0.0.0 -p 3000"
```
Alxzu: Won't this be affected in some way by Dockerfile:51?

juan-apa:
> What do you think about running common tasks when starting the container? such as migrations or db creation?

@Alxzu I thought About this, there are 2 reasons why I didn't set this up:
1. When docker is not being used, the developer will have to run these manually, so there should be no reason not run these using docker.
2. I did not see this done in any of the examples I used to write this image. 

If you think this should be present, I'll gladly add it

Alxzu:
> > What do you think about running common tasks when starting the container? such as migrations or db creation?
> 
> @Alxzu I thought About this, there are 2 reasons why I didn't set this up:
> 
> 1. When docker is not being used, the developer will have to run these manually, so there should be no reason not run these using docker.
> 2. I did not see this done in any of the examples I used to write this image.
> 
> If you think this should be present, I'll gladly add it

No, I think you're right. This is good as it is 👍

docker-compose.yml
juan-apa: What happens in this case, is that the docker-compose's entrypoint overrides the one present in the Dockerfile. Because of this, if the developer wants to change the entrypoint later, there will not be any need to rebuild the image, just changing the entrypoint in the docker-compose will suffice.

Alxzu: Make sense, thanks for the clear answer 👍

Dockerfile
juan-apa: So, the solution to this, was adding the the project's `bin` folder to the `PATH`. Now the commands present in that folder (such as rails, yarn, webpack), are globally accessible.

README.md
```diff
@@ -37,6 +38,13 @@ This template comes with:
 8. `rails s`
 9. You can now try your REST services!
 
+## How to use with docker
+
+1. Have docker and docker-compose installed
+2. build and run the application with `docker-compose up`
```
sebalopez: build and run are actually two commands, `docker-compose build && docker compose up`

juan-apa: I know, but if the image has not been built, `docker-compose up` builds it, and then runs it.

Dockerfile
juanmanuelramallo: I'm not sure about the reminder, it's fine as it is.

Issue #191, opened 2020-03-12
Add Pagy as default pagination gem


jpascual1994:
I'm creating this issue to keep track of the idea of adding [Pagy](https://github.com/ddnexus/pagy) as the default pagination gem.

Feel free to discuss other alternatives like [Kaminari](https://github.com/kaminari/kaminari), [Will Paginate](will_paginate), etc

juanmanuelramallo:
👍 for using Pagy

PR #196, opened 2020-04-08
Add Pagy

jpascual1994:
### Description
Add Pagy as default pagination gem

Solves #191


config/initializers/pagy.rb
```diff
@@ -0,0 +1 @@
+# Pagy initializer file (3.7.5)
```
santib, line 1: shouldn't we have this content instead? https://github.com/ddnexus/pagy/blob/master/lib/config/pagy.rb

README.md
```diff
@@ -64,6 +64,7 @@ This template comes with:
 - [Jbuilder](https://github.com/rails/jbuilder) for json views
 - [Letter Opener](https://github.com/ryanb/letter_opener) for previewing a mail in the browser
 - [Oj](https://github.com/ohler55/oj) for optimized json
+- [Pagy](https://github.com/ddnexus/pagy) for pagination
```
santib, line 67: and tell @jdrosales17 to update his post 😂

config/initializers/pagy.rb
jpascual1994: Yes, I remove the content because Rubocop was complaining about some lines.
If you think it's worth I can add it and exclude the file

README.md
jdrosales17: Damn... now I definitely have to add this gem to the post 😆

config/initializers/pagy.rb
santib: I'd ignore rubocop on this file yes 👍

PR #199, opened 2020-04-14
Add Pundit

fedeagripa:
We don't have any authorization management right now, so we propose to start using pundit for this because we know it works fine and is simple to use & maintain.

If you try to propose `cancan` or `cancancan` please first feel confident that it won't be a `cancancancan` in the near future 😅 .... besides joking, feel free to propose any other but be sure to explain why and pros


app/controllers/application_controller.rb
```diff
@@ -1,4 +1,5 @@
 class ApplicationController < ActionController::Base
+  include Pundit
```
santib, line 2: would it be possible to add these checks?
```
  after_action :verify_authorized, except: :index, unless: :devise_controller?
  after_action :verify_policy_scoped, only: :index
```
https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used

app/controllers/api/v1/users_controller.rb
```diff
@@ -1,6 +1,8 @@
 module Api
   module V1
     class UsersController < Api::V1::ApiController
+      before_action :auth_user
```
santib, line 4: I think you need to do this same thing in others controllers:
`passwords_controller.rb      registrations_controller.rb  sessions_controller.rb       settings_controller.rb`

fedeagripa: I though about it but those only add callbacks and extends `DeviseTokenAuth` what would you auth there?

santib: oh you are right

app/policies/user_policy.rb
```diff
@@ -0,0 +1,13 @@
+class UserPolicy < ApplicationPolicy
+  def show?
+    true
+  end
+
+  def profile?
+    update?
+  end
```
santir489, line 8: Maybe  `alias profile? update?`

fedeagripa: 🤔 is a possibility but they are not conceptually alias imo

app/controllers/application_controller.rb
fedeagripa: Well I liked the intent on this, but i'm having problems because adding that involves ALL controllers in the base, so we should be excluding all controllers besides `user_controller`, and I don't really like it...
opinions?

santib: I strongly recommend adding them. Otherwise people will forget about pundit not call the `authorize` method. You can ignore those controllers if it doesn't make sense to authorize there. For example we have `after_action :verify_authorized, except: :index, unless: :devise_controller?`

santib: not sure if you can use `, unless: :devise_controller?` or something similar

app/policies/application_policy.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def show?
+    false
+  end
+
+  def create?
+    false
+  end
+
+  def new?
```
jdrosales17, line 21: Should we define this despite it won't be used in APIs ?

app/policies/application_policy.rb
```diff
@@ -0,0 +1,49 @@
...
+    create?
+  end
+
+  def update?
+    false
+  end
+
+  def edit?
```
jdrosales17, line 29: Same here.

app/policies/application_policy.rb
fedeagripa: I would say yes, also this is an autogenerated file that is going to be inherited from other policies.
And time to time this base is used to implement webs too

app/controllers/application_controller.rb
santib: @fedeagripa what happens with the settings controller? do you need to exclude it too?

fedeagripa: I ended up excluding the after actions in the controller itself, as that endpoint does not even requiere auth I think there is no need to policies.
Do you think it worth to add a policy specific to return true?

santib: I'm ok with that, I just hope it doesn't set a bad precedent 🙈

PR #202, opened 2020-05-11
Add eager loading gem for basic improvements

fedeagripa:
I found this gem (https://github.com/salsify/goldiloader)  that automatically optimize queries to user eager loading and I think that it would be great to have it as our default, and of course it can be disabled on a specific relationship or on a specific query.

A basic diff of the queries according to doc is:
```
> blogs = Blogs.limit(5).to_a
# SELECT * FROM blogs LIMIT 5

> blogs.each { |blog| blog.posts.to_a }
# SELECT * FROM posts WHERE blog_id = 1
# SELECT * FROM posts WHERE blog_id = 2
# SELECT * FROM posts WHERE blog_id = 3
# SELECT * FROM posts WHERE blog_id = 4
# SELECT * FROM posts WHERE blog_id = 5
```

Ends up being:

```
> blogs = Blogs.limit(5).to_a
# SELECT * FROM blogs LIMIT 5

> blogs.each { |blog| blog.posts.to_a }
# SELECT * FROM posts WHERE blog_id IN (1,2,3,4,5)
```


juanmanuelramallo:
I think I'd rather have the code improved, instead of adding this gem. I think that having automatic eager loading can be problematic when working with Single Table Inheritance, or polymorphism, or active storage attachments, but I haven't used this gem before, so I'm not entirely sure.

If we want to do:
```ruby
blogs.each { |blog| blog.posts.to_a }
``` 

we can rewrite it as: 
```ruby
blogs.include(:posts).each { |blog| blog.posts.to_a }
``` 

so that we can still hit the DB only once, and have all the information we need. Same result.

We already have bullet configured to notify us these kind of queries I believe.

fedeagripa:
@juanmanuelramallo Ignore the ugly example, I agree we can easily write it in a better way. My aim (as discussed in slack thread) here is to solve all the warnings bullet generates and are omitted or forgotten (or even worse... that we don't have time to fix).

juanmanuelramallo:
I still believe we would be shadowing too much of the SQL generation. It's like adding one level of indirection to generate SQL, where ActiveRecord is already powerful enough to solve it by adding an `includes` statement. 

Anyways I don't have a strong opinion on this, so feel free to move it forward.

fedeagripa:
I come from a project where people added an ugly banner in the main page showing the forgotten eager loads 👀  and I've seen a couple mvps forgetting this, so I see a big win for this change. But again, don't want to force this, lets try to conclude if we prefer bullet as a soft solution instead of this as a hard one

santib:
I'm with @juanmanuelramallo on this one. If someone wants to add it in a project I'm ok, but I'm not completely sold on having this on a base project.

Personally, I prefer the approach of being notified about these problems (Bullet) instead of trying to automagically fix them. The question I ask myself is: if this is a good approach, why wouldn't have Rails done it?

fedeagripa:
I buy the Rails question 💸
I will keep this as a solution for some specific projects instead

PR #204, opened 2020-05-14
Remove Draper gem

### LINKED ISSUE #203: Remove draper gem
As discussed, draper adds complexity that we don't really need. We'll replace our current presenter by making a PORO that inherits from a `BasePresenter`. We should also add some documentation to this `BasePresenter` on when and how to use presenters, maybe by linking to a place where this is well explained.

brunvez:
Closes #203


app/models/user.rb
```diff
@@ -41,12 +41,6 @@ class User < ApplicationRecord
 
   before_validation :init_uid
 
-  def full_name
-    return username if first_name.blank?
-
-    "#{first_name} #{last_name}"
-  end
```
santib, line 48: ❤️

spec/rails_helper.rb
```diff
@@ -2,7 +2,10 @@
 ENV['RAILS_ENV'] ||= 'test'
 require 'simplecov'
 
-SimpleCov.start do
+SimpleCov.start 'rails' do
+  add_group 'Forms', 'app/forms'
+  add_group 'Policies', 'app/policies'
+  add_group 'Presenters', 'app/presenters'
```
santib, line 8: 👏

Gemfile
```diff
@@ -11,7 +11,6 @@ gem 'bootsnap', '~> 1.4', '>= 1.4.5'
 gem 'delayed_job_active_record', '~> 4.1', '>= 4.1.4'
 gem 'devise', '~> 4.7.1'
 gem 'devise_token_auth', '~> 1.1', '>= 1.1.3'
-gem 'draper', '~> 3.1'
```
santib, line 14: I think you need to run `bundle install` so the `Gemfile.lock` gets updated

santib: oh and also remove the reference in the `README`

martinjaimem:
Niceee 😎

app/controllers/api/v1/users_controller.rb
```diff
@@ -3,14 +3,18 @@ module V1
     class UsersController < Api::V1::ApiController
       before_action :auth_user
 
-      def show; end
+      def show
+        @user = UserPresenter.new(current_user)
```
fedeagripa, line 7: what about moving the `UserPresenter.new(current_user)` to a private method as explained in this post: https://blog.rstankov.com/rails-anti-pattern-setting-view-variables-in-before-actions/ ?
Just to be a bit more DRY

brunvez:
As per https://github.com/rootstrap/rails_api_base/pull/209, on hold until AA supports PORO presenters or we find an easy way to get around that.

app/policies/admin_user_policy.rb
```diff
@@ -0,0 +1,25 @@
+class AdminUserPolicy < ApplicationPolicyq
```
santib, line 1:
```suggestion
class AdminUserPolicy < ApplicationPolicy
```

sebastiancaraballo:
@santib Should we close this?

santib:
> @santib Should we close this?

sure, let's close it and whenever we have this discussion we can re-open it if we want 👍

PR #209, opened 2020-05-24
Make PORO presenter work with ActiveAdmin

santib:
This is a POC of a PORO presenter working with ActiveAdmin. At first I tried to do it without using the gem but it was really really hard. The gem made it work but I think it has some issues for example with pagination, sorting, etc.

Seems like initially ActiveAdmin wanted to support different presenter/decorator implementations but at some point they decided to stick with Draper and not officially support any other alternative. https://github.com/activeadmin/activeadmin/issues/3600#issuecomment-62959836 

I think the presenter pattern is necessary in ActiveAdmin to prevent presentational methods leak into the models, so that leads us to the question, do we feel confident going against the path that ActiveAdmin has taken? Should we replace Draper for PORO presenters? Should we use a gem which last update was on 2015? Is it worth it?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'rails', '~> 6.0.0'
 
 gem 'active_storage_base64', '~> 1.0.0'
 gem 'activeadmin', '~> 2.4'
+gem 'activeadmin-poro-decorator', '~> 0.2'
 gem 'aws-sdk-s3', '~> 1', require: false
 gem 'bootsnap', '~> 1.4', '>= 1.4.5'
 gem 'delayed_job_active_record', '~> 4.1', '>= 4.1.4'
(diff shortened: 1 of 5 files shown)


brunvez:
This all started because we wanted to remove Draper. One of the motivations of such action was to have fewer dependencies in our Gemfile. Keeping that in mind I think we'd rather have Draper, a well known and maintained gem, rather than an outdated one.

As you said implementing a PORO presenter that's compatible with AA isn't easy and we'd end up with similar code to Draper. IMO the best course of action is to fix the problem from the root and try to submit a PR to AA to allow supporting what we need. In the meantime, we can leave the PR to remove Draper on hold. At least that's my take on this.

santib:
> This all started because we wanted to remove Draper. One of the motivations of such action was to have fewer dependencies in our Gemfile. Keeping that in mind I think we'd rather have Draper, a well known and maintained gem, rather than an outdated one.
> 
> As you said implementing a PORO presenter that's compatible with AA isn't easy and we'd end up with similar code to Draper. IMO the best course of action is to fix the problem from the root and try to submit a PR to AA to allow supporting what we need. In the meantime, we can leave the PR to remove Draper on hold. At least that's my take on this.

@brunvez I completely agree 💯

PR #211, opened 2020-05-26
Fix Admin not working

brunvez:
## Issue

Admin was not working. Now it works. Because it's fixed.

## Reason

Pundit was missing ActiveAdmin configuration


brunvez:
> Do you intended to add `binding_of_caller` or is a leftover from your dark sorcery? If so, could you point me where is needed exactly?

Oh, yeah. That's a gem so that BetterError works better. It gives you a console right on the browser with the current execution context

brunvez:
@fedeagripa here's how it looks

image

app/policies/application_policy.rb
```diff
@@ -7,31 +7,31 @@ def initialize(user, record)
   end
 
   def index?
-    false
+    user.is_a?(AdminUser)
```
santib, line 10: maybe we don't need these changes in `app/policies/application_policy.rb`

I'd have `app/policies/admin/application_policy.rb` define the policies for `AdminUser`

thoughts?

santib: I say it because `User` and `AdminUser` are different models, so I think policies should be different as well and would be easier to manage

Gemfile
```diff
@@ -43,6 +43,7 @@ end
 group :development do
   gem 'annotate', '~> 3.0', '>= 3.0.3'
   gem 'better_errors', '~> 2.5', '>= 2.5.1'
+  gem 'binding_of_caller', '~> 0.8.0'
```
santib, line 46: why is this needed?

app/policies/application_policy.rb
brunvez: That'd be an improvement but Pundit reflects on policies using the model, so it uses the same policies as we normally use. To change this we would need to implement some way of getting policies from the `Admin` namespace if it's for Active Admin.

I completely agree with the comment but I'd add it as an enhancement issue.

Gemfile
brunvez: It's just so we can use the browser console described here https://github.com/rootstrap/rails_api_base/pull/211#issuecomment-634264205

app/policies/application_policy.rb
brunvez: @santib I created the issue to deal with that, but I think this should be merged ASAP https://github.com/rootstrap/rails_api_base/issues/212

santib: I agree 💯

PR #216, opened 2020-05-29
adding arctic admin and moving js and stylesheets to webpacker

TimoPeraza:
### Issue

Active admin is not responsive. It doesn't look fine in mobiles.

### Reason

We chose arctic_admin because adds the responsive feature, which we need, and gives a better look to the admin page. Also, nowadays, arctic admin has support.


Alxzu:
Please include at least some pictures into the description to see how it looks like on both, desktop and mobile 🙏

PR #222, opened 2020-06-16
Move tests down in the pipeline

juan-apa:
#### Board:
* N/A
---
#### Description:
* This PR moves the `code_analysis` task before running the tests in the circleci pipeline. This is because the `code_analysis` takes less time to execute than the tests. This can potentially save computing time in circleci to allow other workflows to run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,10 +37,6 @@ jobs:
           name: I18n Health
           command: bundle exec i18n-tasks health
 
-      - run:
-          name: run tests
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -52,6 +48,11 @@ jobs:
           path: /tmp/test-results
           destination: test-results
 
+      # run tests
```
juanmanuelramallo, line 51: Oh thanks I had no idea this was to run the tests 🙈🙈🙈🙈🙈🙈🙈🙈

Issue #226, opened 2020-06-24
Update code_analysis task to run analysis on the whole codebase


juanmanuelramallo:
## Why
Because we want to run Rubocop, Reek (any other analysis) on the whole codebase, not just on specific folders; so that we enforce best practices across all the codebase

## What
Update our code_analysis rake task to run the analysis on all the files, in other words in the root directory

juan-apa:
@juanmanuelramallo can this issue be closed now?

juanmanuelramallo:
Yes!

Issue #225, opened 2020-06-24
Fix existing reek issues


juanmanuelramallo:
## Why
There are reek issues in the current state of the codebase. Fix those so that new projects can have a clean rails app to start development. 

## What
Run `bundle exec reek` and fix the errors encountered

juan-apa:
can this issue be closed?

PR #228, opened 2020-06-26
Expand code analysis tasks scope

juan-apa:
#### Issues:
* #225 
* #226 
---
#### Description:
- Extends the scope of the code_analysis task to the whole codebase.
---
#### Notes:
* There is only one excluded folder for reek, which is db/migrate.


lib/tasks/code_analysis.rake
```diff
@@ -1,6 +1,6 @@
-task :code_analysis do
+task code_analysis: :environment do
   sh 'bundle exec brakeman . -z -q'
-  sh 'bundle exec rubocop app config lib spec'
-  sh 'bundle exec reek app config lib'
+  sh 'bundle exec rubocop .'
+  sh 'bundle exec reek app lib public spec tmp'
```
santib, line 4: why not `bundle exec reek .` ?

.rubocop.yml
```diff
@@ -27,3 +38,27 @@ Metrics/LineLength:
 Rails/Delegate:
   Exclude:
     - app/controllers/api/concerns/act_as_api_request.rb
+
+Style/Documentation:
```
santib, line 42: I'd like to keep them in order
```
Rails/X

Style/X

etc
```

Gemfile.lock
```diff
@@ -75,7 +75,7 @@ GEM
       zeitwerk (~> 2.2, >= 2.2.2)
     addressable (2.7.0)
       public_suffix (>= 2.0.2, < 5.0)
-    annotate (3.0.3)
+    annotate (3.1.1)
```
santib, line 78: why were gems updated?

db/migrate/20200128183708_create_settings.rb
```diff
@@ -5,6 +5,8 @@ def change
       t.string :value
 
       t.timestamps
+
+      t.index :key, unique: true
```
santib, line 9: why was this index created?

lib/tasks/code_analysis.rake
juan-apa: that actually broke circleCI. If reek used the `.` scope, it actually started analyzing the gems. Here you can see an example run: https://app.circleci.com/pipelines/github/rootstrap/rails_api_base/75/workflows/ddb79150-954d-46d3-9364-edb7baa6d2ea/jobs/747

db/migrate/20200128183708_create_settings.rb
juan-apa: this was created because rubocop now fails if there is a `unique` validation in the model and no `unique index` in the migration

Gemfile.lock
juan-apa: I actually ran `bundle update` while trying to diagnose why circleci was failing to run the `code_analysis` task (spoiler, it wasn't related). Left it updated because I thought it would do no harm. I can roll this update back if you think this should be done in another PR or not done at all right now.

lib/tasks/code_analysis.rake
juan-apa: when running `bundle exec reek .` in the local env, there are no problems. The problems occur only in circleCI. Maybe this should be addressed before merging this PR by changing `.circleci/config.yml`. What do you think?

Gemfile.lock
santib: I usually prefer having PRs tackle a specific thing and not doing more than that, but I can make an exception since I agree it does no harm 😄

.rubocop.yml
```diff
@@ -13,6 +13,17 @@ AllCops:
...
+Layout/ClassStructure:
+  Enabled: true
+
+Layout/SpaceBeforeFirstArg:
+  Exclude:
+    - app/views/api/**/**/*
+
+Lint/AmbiguousBlockAssociation:
+  Exclude:
+    - spec/**/*
```
juanmanuelramallo, line 25: Why these new config? Aren't these already provided by rubocop-rootstrap?

lib/tasks/code_analysis.rake
juanmanuelramallo: The issue there is that reek was run against the bundled dependencies, I guess it's happening the same with rubocop. 

So we should exclude the path where our gems are bundled (in circleCI)

juanmanuelramallo (approved):

jpascual1994:
@juan-apa Is it ready to merge? Could you solve the conflicts?

jpascual1994 (approved):
🚀

PR #229, opened 2020-07-01
Improve dockerfile

juan-apa:
#### Description:
* This PR improves the docker support. What happened before, was that  when developing inside the container, the following problems occured:
   1. The container would start using the bundler version 1.x
   2. Gems would stay outdated in some cases
* Also This PR makes the final image smaller. I took some ideas from [thoughtbot docker examples](https://thoughtbot.com/blog/rails-on-docker) to make the improvements.
---
### Notes:
* To execute the rails commands, there are 2 ways:
    1. `bin/rails ...`
    2. `bundle exec rails ...`
* Directly calling `rails` uses the bundler version 1.x and an error message will appear, signaling that a bundler version 2 or greater is needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ruby:2.6.3
 
 RUN apt-get update -qq && \
-    apt-get install -y build-essential openssl libssl-dev nodejs libpq-dev less vim nano libsasl2-dev
+    apt-get install -y build-essential libssl-dev nodejs libpq-dev less vim nano libsasl2-dev
 
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
 RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,50 +1,35 @@
 FROM ruby:2.6.3
 
-LABEL Name=rails_api_base Version=1.0
+RUN apt-get update -qq && \
+    apt-get install -y build-essential openssl libssl-dev nodejs libpq-dev less vim nano libsasl2-dev
 
-ENV PATH=/app/bin:$PATH
-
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,50 +1,35 @@
 FROM ruby:2.6.3
 
-LABEL Name=rails_api_base Version=1.0
+RUN apt-get update -qq && \
+    apt-get install -y build-essential openssl libssl-dev nodejs libpq-dev less vim nano libsasl2-dev
```
sebalopez, line 4: just curious, what is the advantage of libpq-dev over postgresql-client ?

juan-apa: Actually I changed it, because thoughtbot dockerfile was using it. After doing some research I found out that `libpq-dev` should have a smaller footprint than `postgresql-client`.
[Here](https://www.postgresql.org/download/linux/ubuntu/) you can check it out.

juan-apa:
> Great job!
> I don't fully understand the Ruby/Node dependency management particular but I like the idea of not mounting the host node_modules and gems, and removing the Heroku CLI.
> Perhaps we should add some comment stating that this image is not meant for Production use as-is given it includes several utilities that would allow unneeded external access.

The thing is that rails projects use nodejs to handle js dependencies. By default rails uses yarn to handle those. What do you suggest to prevent pulling yarn with wget?

juan-apa:
> I wonder if we should include DB migrations (`bundle exec rails db:migrate`) as part of the entrypoint, or whether it should be executed externally.
> I am doing it for production deployments where the database is assumed to exist and be ready before container, not sure if it should be the same for development or if you could also include `db:create`

I'm open to suggestions, I didn't include database creation/migrations because of what you what you said (in production, the database is usually already created).
I've been using this image mainly to develop. If that's the main use case, then I think migrations should not be included in the Dockerfile, because I could be working in a migration, and the next time I open the container, the migration would be run (maybe without me noticing it)

If the purpose is to use this image for production, then maybe it should run the database creation and migration tasks.  When running `rails db:create`, if the database exists, then no error is thrown, it exits the process successfully.

PR #235, opened 2020-07-21
improve coverage

fedeagripa:
Improve code coverage

### DIFF
--- a/spec/decorators/user_decorator_spec.rb
+++ b/spec/decorators/user_decorator_spec.rb
@@ -1,0 +1,8 @@
+describe UserDecorator do
+  let(:user) { create(:user) }
+  let(:decorated_user) { user.decorate }
+
+  it 'returns same email' do
+    expect(decorated_user.email).to eq(user.email)
+  end
+end
--- a/spec/mailers/application_mailer_spec.rb
+++ b/spec/mailers/application_mailer_spec.rb
@@ -1,0 +1,13 @@
+module MockMailer
+  class UserMailer < ApplicationMailer
+    def test_email; end
+  end
+end
+
+describe ApplicationMailer do
+  it 'job is created' do
+    expect {
+      MockMailer::UserMailer.test_email.deliver_later
+    }.to have_enqueued_job.on_queue('mailers')
+  end
+end
--- a/spec/policies/admin_user_policy_spec.rb
+++ b/spec/policies/admin_user_policy_spec.rb
@@ -1,0 +1,11 @@
+describe AdminUserPolicy do
+  subject { described_class }
+
+  permissions :update?, :index?, :show?, :create?, :new?, :edit?, :destroy? do
+    let(:admin) { create(:admin_user) }
+
+    it 'allows access ' do
+      expect(subject).to permit(admin, admin)
+    end
+  end
+end
--- a/spec/factories/admin_users.rb
+++ b/spec/factories/admin_users.rb
@@ -1,0 +1,6 @@
+FactoryBot.define do
+  factory :admin_user do
+    email    { Faker::Internet.unique.email }
+    password { Faker::Internet.password(min_length: 8) }
+  end
+end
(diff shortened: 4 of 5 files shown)


spec/decorators/user_decorator_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
```
martinjaimem, line 1: I think we can remove this, we already have the `,rspec` file

fedeagripa: good catch 💪

spec/mailers/application_mailer_spec.rb
```diff
@@ -0,0 +1,12 @@
+class UserMailer < ApplicationMailer
+  include ActiveJob::TestHelper
+  def test_email; end
+end
```
brunvez, line 4: Just thinking allow, what if we put this in the `support` folder just to show how a mock could be done? I usually prefer them close to the code too so doubtful about it

spec/policies/admin_user_policy_spec.rb
```diff
@@ -0,0 +1,11 @@
+describe AdminUserPolicy do
+  subject { described_class }
+
+  permissions :update?, :index?, :show?, :create?, :new?, :edit?, :destroy? do
+    let(:admin) { create(:admin_user) }
+
+    it 'allow access ' do
```
brunvez, line 7:
```suggestion
    it 'allows access' do
```

.codeclimate.yml
```diff
@@ -57,3 +57,4 @@ plugins:
       - spec/**/*
 exclude_patterns:
   - 'babel.config.js'
+  - 'app/jobs/application_job.rb'
```
santib, line 60: maybe it's better to not exclude it, otherwise if projects adds custom logic in this file they won't be running code climate on that code

spec/decorators/user_decorator_spec.rb
```diff
@@ -0,0 +1,8 @@
+describe UserDecorator do
+  let(:user) { create(:user) }
+  let(:decorated_user) { user.decorate }
+
+  it 'returns the displayed days' do
```
santib, line 5: displayed days?

spec/mailers/application_mailer_spec.rb
```diff
@@ -0,0 +1,12 @@
+class UserMailer < ApplicationMailer
+  include ActiveJob::TestHelper
```
santib, line 2: is this necessary? shouldn't it be incldued in `spec/rails_helper.rb` ?

spec/mailers/application_mailer_spec.rb
```diff
@@ -0,0 +1,12 @@
+class UserMailer < ApplicationMailer
```
santib, line 1: maybe we can namespace it, otherwise it could conflict with a real `UserMailer`

spec/mailers/application_mailer_spec.rb
fedeagripa: seems to be necessary, I can add it there yes 💪

.codeclimate.yml
fedeagripa: agree, but I don't see any way to test this apart from adding any silly job, do you think it worth it?

spec/mailers/application_mailer_spec.rb
fedeagripa: I would prefer this to be deleted after some real tests.... so I prefer to keep it here and developers deleting it later, sounds good?

spec/policies/admin_user_policy_spec.rb
santib: @fedeagripa this isn't fixed yet

fedeagripa: fixed now, thanks!

PR #243, opened 2020-08-12
Add bootstrap script

brunvez:
#### Description:

This PR adds a bootstrap script as both a place to set up projects from scratch quicker and for teams to, later on, add custom configuration steps.


bootstrap.sh
```diff
@@ -0,0 +1,27 @@
...
+project_name=$1
+
+# install ruby gems
+bundle install
+
+#install node packages
+yarn
+
+# copy secrets from the example adding a new application secret
+application_secret=$(bundle exec rake secret)
```
santib, line 20: I think this is not necessary, rails uses a tmp file for dev env secrets

`railties/lib/rails/application.rb`
```ruby
    # The secret_key_base is used as the input secret to the application's key generator, which in turn
    # is used to create all MessageVerifiers/MessageEncryptors, including the ones that sign and encrypt cookies.
    #
    # In development and test, this is randomly generated and stored in a
    # temporary file in <tt>tmp/development_secret.txt</tt>.
    #
    # In all other environments, we look for it first in ENV["SECRET_KEY_BASE"],
    # then credentials.secret_key_base, and finally secrets.secret_key_base. For most applications,
    # the correct place to store it is in the encrypted credentials file.
    def secret_key_base
      if Rails.env.development? || Rails.env.test?
        secrets.secret_key_base ||= generate_development_secret
      else
        validate_secret_key_base(
          ENV["SECRET_KEY_BASE"] || credentials.secret_key_base || secrets.secret_key_base
        )
      end
    end
```

config/database.yml.example
```diff
@@ -10,12 +10,11 @@ default: &default
...
- database:
+ database: sample_project_development
 
 test:
  <<: *default
- database:
+ database: sample_project_test
 
 production:
- <<: *default
```
santib, line 20: shouldn't this be kept for general configs like adapter, encoding, pool, etc?

bootstrap.sh
```diff
@@ -0,0 +1,27 @@
...
+sed "s/SECRET_KEY_BASE:/SECRET_KEY_BASE: ${application_secret}/g" config/application.yml.example > config/application.yml
+
+# copy database configuration and change the project name with the given one
+sed "s/sample_project/${project_name}/g" config/database.yml.example > config/database.yml
+
+# setup database
+bundle exec rails db:create db:migrate db:seed
```
juanmanuelramallo, line 27:
```suggestion
bundle exec rails db:setup
```
Does exactly the same

bootstrap.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+set -exu
+
+die () {
+    echo >&2 "$@"
+    exit 1
+}
+
+[ "$#" -eq 1 ] || die "1 argument required, $# provided. Pleas add the project name"
```
juanmanuelramallo, line 9: Lovely bash script, pleas fix typo 🙈

bootstrap.sh
brunvez: Yeah but instead of running migrations, it loads the schema. My concern is that if migrations break no one will realize

config/database.yml.example
brunvez: I think those are managed on the URL itself or as config variables which Rails recognizes

bootstrap.sh
brunvez: That makes it much easier, won't there be any problems with deleting the `tmp` folder? I know I do it once in a while so I'm afraid things signed before might be invalid

config/database.yml.example
santib: I think not everything is present in the URL, for instance: encoding, pool, any probably others

bootstrap.sh
santib: IDK, I guess it's not a problem in dev environment if the session gets expired or something like that

brunvez: I'll leave it, I don't want to find side effects and if anything it serves as a way to see how to generate dynamic configs

santib: but isn't the `SECRET_KEY_BASE` env var completely ignored in dev env?

santib: I run the script and..

image

brunvez: lol, so let's just remove `SECRET_KEY_BASE:` from `application.yml.example`? I believe it throws an error if I don't set a value

santib: Seems like the ENV VAR is directly used here: `config.secret_key_base = ENV['SECRET_KEY_BASE']` in `config/application.rb` .. looks weird to me 😓

PR #245, opened 2020-08-19
Add auto run for Rspec

brunvez:
#### Description:

This allows running specs for the changed files by following some conventions. It's really handy on big projects TBH. The goal of this PR is to open a discussion, if approved we should ask permission from the author to add it.

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/bin/auto_rspec
+++ b/bin/auto_rspec
@@ -1,0 +1,67 @@
+#!/bin/bash
+#
+# Automatically find and run specs corresponding to new and changed files or
+# terms passed in.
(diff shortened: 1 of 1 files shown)


fedeagripa:
Would you mind provide some output examples (or at least one) ?

brunvez:
> Would you mind provide some output examples (or at least one) ?

Seems fair, this is how you'd run it and the output you'd get

image

AdminUser didn't have any specs so no tests are run

brunvez:
> Here, take my money 💰
> Also, I'm assuming this is something we need to add to our pre-push process, right?

😂 I think it's a good idea to add it to the pre-push hook I don't know what others think

santib (approved):

PR #248, opened 2020-08-25
Feature/improve tests

JuannFerrari:
### Description

This PR tries to improve the code coverage of our api_base and remove some unused code.

---
#### Notes:
- I have deleted the following rescue sentences because we can't rescue these (since Rails 3 [#ref)](https://github.com/rails/rails/issues/671)
```
      rescue_from ActionController::RoutingError,      with: :render_not_found
      rescue_from AbstractController::ActionNotFound,  with: :render_not_found
```

- from `models/user.rb` we had this unused method

```
  def uses_email?
    provider == 'email' || email.present?
  end
```

in the `Api::V1::ApiController` we have the next method:

```      
     def render_error(exception)
        raise exception if Rails.env.test?

        # To properly handle RecordNotFound errors in views
        return render_not_found(exception) if exception.cause.is_a?(ActiveRecord::RecordNotFound)

        logger.error(exception) # Report to your error managment tool here

        return if performed?

        render json: { error: I18n.t('api.errors.server') }, status: :internal_server_error
      end
```

This line `raise exception if Rails.env.test?` prevent us to test the `render_error` method in a proper way and increase the coverage, WDYT?

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -43,7 +43,7 @@ class ApplicationPolicy
     end
 
     def resolve
-      scope.all
+      scope.all if user.is_a?(AdminUser)
     end
   end
 end
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -43,7 +43,7 @@ class ApplicationPolicy
     end
 
     def resolve
-      scope.all if user.is_a?(AdminUser)
+      scope.all
     end
   end
 end
--- a/spec/policies/active_admin/page_policy_spec.rb
+++ b/spec/policies/active_admin/page_policy_spec.rb
@@ -2,7 +2,7 @@ describe ActiveAdmin::PagePolicy do
   subject { described_class }
 
   permissions :show? do
-    let!(:user) { create(:user) }
+    let(:user) { create(:user) }
     let(:record) { double('DashboardRecord', name: 'Dashboard') }
 
     it 'allow access if record.name is Dashboard' do
--- a/spec/policies/active_admin/page_policy_spec.rb
+++ b/spec/policies/active_admin/page_policy_spec.rb
@@ -3,7 +3,7 @@ describe ActiveAdmin::PagePolicy do
 
   permissions :show? do
     let!(:user) { create(:user) }
-    let(:record) { OpenStruct.new(name: 'Dashboard') }
+    let(:record) { double('DashboardRecord', name: 'Dashboard') }
 
     it 'allow access if record.name is Dashboard' do
       expect(subject).to permit(user, record)
--- a/spec/policies/active_admin/page_policy_spec.rb
+++ b/spec/policies/active_admin/page_policy_spec.rb
@@ -10,7 +10,7 @@ describe ActiveAdmin::PagePolicy do
     end
 
     context 'when the record name is distinct to Dashboard' do
-      let(:record) { OpenStruct.new(name: 'not-valid') }
+      let(:record) { double('NoneDashboardRecord', name: 'not-valid') }
 
       it 'denies access' do
         expect(subject).not_to permit(user, record)
--- a/spec/policies/active_admin/page_policy_spec.rb
+++ b/spec/policies/active_admin/page_policy_spec.rb
@@ -1,0 +1,20 @@
+describe ActiveAdmin::PagePolicy do
+  subject { described_class }
+
+  permissions :show? do
+    let!(:user) { create(:user) }
+    let(:record) { OpenStruct.new(name: 'Dashboard') }
+
+    it 'allow access if record.name is Dashboard' do
+      expect(subject).to permit(user, record)
+    end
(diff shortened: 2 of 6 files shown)


horacio:
Neat, man!

> This line `raise exception if Rails.env.test?` prevent us to test the `render_error` method in a proper way and increase the coverage, WDYT?

Just an idea: can we use something like `allow(Rails).to receive(:env).and_return 'production'.inquiry` around that test?

JuannFerrari:
@horacio love the idea! I will try to implement the test in the next commit

app/policies/application_policy.rb
```diff
@@ -43,7 +43,7 @@ def initialize(user, scope)
     end
 
     def resolve
-      scope.all
+      scope.all if user.is_a?(AdminUser)
```
brunvez, line 46: What happens if the user is not an admin? This is used across more places and I think the default is alright, it'd just be overridden in other policies if you wish to limit what the user can see.

We are also violating LSP, we should at least return `scope.none` to not break that principle. But I think it still makes more sense to return all records.

spec/policies/active_admin/page_policy_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe ActiveAdmin::PagePolicy do
+  subject { described_class }
+
+  permissions :show? do
+    let!(:user) { create(:user) }
+    let(:record) { OpenStruct.new(name: 'Dashboard') }
```
brunvez, line 6:
```suggestion
    let(:record) { double('DashboardRecord', name: 'Dashboard') }
```

spec/policies/active_admin/page_policy_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe ActiveAdmin::PagePolicy do
+  subject { described_class }
+
+  permissions :show? do
+    let!(:user) { create(:user) }
```
brunvez, line 5:
```suggestion
    let(:user) { create(:user) }
```

spec/policies/active_admin/page_policy_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    let(:record) { OpenStruct.new(name: 'Dashboard') }
+
+    it 'allow access if record.name is Dashboard' do
+      expect(subject).to permit(user, record)
+    end
+
+    context 'when the record name is distinct to Dashboard' do
+      let(:record) { OpenStruct.new(name: 'not-valid') }
```
brunvez, line 13:
```suggestion
      let(:record) { double('NoneDashboardRecord', name: 'not-valid') }
```

brunvez:
> Neat, man!
> 
> > This line `raise exception if Rails.env.test?` prevent us to test the `render_error` method in a proper way and increase the coverage, WDYT?
> 
> Just an idea: can we use something like `allow(Rails).to receive(:env).and_return 'production'.inquiry` around that test?

I think we could just get rid of that. There's an option to raise exceptions on tests I think it was something like `config.action_dispatch.show_exceptions` which we should be able to deactivate just for that test

JuannFerrari:
@brunvez Thanks for your feedback! I was looking to disable that config but it seems to have already been disabled in this file  https://github.com/rootstrap/rails_api_base/blob/eb7ce016b3c9d1448b7abfdc6b7899b0ea65f236/config/environments/test.rb#L27

So I think that the approach you propose is not possible 🤔 , I took the suggestion of @horacio to do the test. Anyways I'm open to change it

spec/policies/application_policy_spec.rb
```diff
@@ -2,10 +2,30 @@
   subject { described_class }
 
   permissions :update?, :index?, :show?, :create?, :new?, :edit?, :destroy? do
+    context 'when is an user' do
+      let!(:user) { create(:user) }
```
brunvez, line 6: We can just `build` here
```suggestion
      let(:user) { build(:user) }
```

spec/policies/application_policy_spec.rb
```diff
@@ -2,10 +2,30 @@
...
+
+      it 'denies access' do
+        expect(subject).not_to permit(user, user)
+      end
+    end
+
+    context 'when is an admin' do
+      let!(:admin) { create(:admin_user) }
```
brunvez, line 14:
```suggestion
      let(:admin) { build(:admin_user) }
```

spec/policies/application_policy_spec.rb
```diff
@@ -2,10 +2,30 @@
...
+
+      it 'permits access' do
+        expect(subject).to permit(admin, admin)
+      end
+    end
+  end
+
+  describe 'scope' do
     let(:user) { create(:user) }
+    let(:mock_model) { OpenStruct.new(all: true) }
```
brunvez, line 24:
```suggestion
    let(:mock_model) { mock('MockModel', all: true) }
```

PR #250, opened 2020-08-31
Check Content-Type header only when a payload is sent

juanmanuelramallo:
Clients use the Content-Type header to tell the server what type of data
has been sent. Hence there is no need for the server to disallow GET or
HEAD requests. More info on how the Content-Type header is/should-be used for
requests [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#:~:text=the%20client%20tells%20the%20server%20what%20type%20of%20data%20is%20actually%20sent) and [here](https://tools.ietf.org/html/rfc7231#section-3.1.1.5).

Updated status_spec to match this behavior.

### DIFF
--- a/app/controllers/api/concerns/act_as_api_request.rb
+++ b/app/controllers/api/concerns/act_as_api_request.rb
@@ -10,8 +10,7 @@ module Api
       end
 
       def check_json_request
-        return unless request_with_body?
-        return if request_content_type.match?(/json/)
+        return if !request_with_body? || request_content_type.match?(/json/)
 
         render json: { error: I18n.t('api.errors.invalid_content_type') }, status: :not_acceptable
       end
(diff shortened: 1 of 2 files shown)


app/controllers/api/concerns/act_as_api_request.rb
```diff
@@ -10,7 +10,8 @@ module ActAsApiRequest
       end
 
       def check_json_request
-        return if request_content_type&.match?(/json/)
+        return unless request_with_body?
```
sebaherrera07, line 13: Personally, I would find it easier to follow if it just had 1 early return.
Maybe something like:
```
return if request_without_body? || request_content_type.match?(/json/)
```

juanmanuelramallo: Nice, looks good

mrubi-rootstrap: I think that what makes it hard to understand logicals expression is the use of both `if` and `unless` in the same method. I try not to use `unless` for that reason. I prefer the condition check to always be for the positive and negate the statement if necessary

jpascual1994 (approved):

PR #253, opened 2020-09-05
Using Rails in api_only mode

juanmanuelramallo:
Api only apps won't need all controller modules used in normal applications,
for instance: ViewPaths, Layouts, Templates, and so on. Moreover, api only apps
won't need all middlewares used in normal applications, for instance:
MethodOverride (used to override the HTTP verb when passing a hidden
field in form submissions), ContentSecurityPolicy (which is a browser-only thing), and so.

One consideration: Devise controllers will always
inherit from ApplicationController, hence we still need to use the
ActAsApiRequest concern in Devise controllers. This is because Devise is shared
between the API-only app and the admin panel.

_Branched off from #250_

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,3 +27,7 @@ Layout/LineLength:
 Rails/Delegate:
   Exclude:
     - app/controllers/api/concerns/act_as_api_request.rb
+
+Rails/LexicallyScopedActionFilter:
+  Exclude:
+    - app/controllers/api/v1/api_controller.rb
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -1,14 +1,14 @@
 module Api
   module V1
-    class ApiController < ApplicationController
-      include Api::Concerns::ActAsApiRequest
+    class ApiController < ActionController::API
+      include Pundit
       include DeviseTokenAuth::Concerns::SetUserByToken
+
+      after_action :verify_authorized, except: :index
+      after_action :verify_policy_scoped, only: :index
 
       before_action :authenticate_user!, except: :status
       skip_after_action :verify_authorized, only: :status
-
-      layout false
-      respond_to :json
 
       rescue_from ActiveRecord::RecordNotFound,        with: :render_not_found
       rescue_from ActiveRecord::RecordInvalid,         with: :render_record_invalid
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   include Pundit
+
   after_action :verify_authorized,
                except: :index,
                unless: -> { :devise_controller? || :active_admin_controller? }
@@ -7,7 +8,6 @@ class ApplicationController < ActionController::Base
                only: :index,
                unless: -> { :devise_controller? || :active_admin_controller? }
   # Prevent CSRF attacks by raising an exception.
-  # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
   def active_admin_controller?
--- a/config/application.rb
+++ b/config/application.rb
@@ -35,5 +35,15 @@ module App
       g.test_framework :rspec
       g.fixture_replacement :factory_bot, dir: 'spec/factories'
     end
+
+    # Only loads a smaller set of middleware suitable for API only apps.
+    # Middleware like session, flash, cookies can be added back manually.
+    # Skip views, helpers and assets when generating a new resource.
+    config.api_only = true
+
+    # ActiveAdmin needs the following middlewares to work properly.
+    config.middleware.use ActionDispatch::Cookies
+    config.middleware.use ActionDispatch::Session::CookieStore
+    config.middleware.use ActionDispatch::Flash
   end
 end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -75,4 +75,6 @@ Rails.application.configure do
     Bullet.rails_logger = true
     Bullet.add_footer = true
   end
+
+  config.debug_exception_response_format = :default
 end
--- a/spec/requests/api/v1/settings_spec.rb
+++ b/spec/requests/api/v1/settings_spec.rb
@@ -6,7 +6,7 @@ describe 'GET api/v1/settings/must_update', type: :request do
       device_version: '1.0'
     }
   end
-  subject { get must_update_api_v1_settings_path, params: params, as: :json }
+  subject { get must_update_api_v1_settings_path, params: params }
 
   context 'with correct settings' do
     let!(:setting_version) { create(:setting_version) }


.rubocop.yml
```diff
@@ -27,3 +27,7 @@ Layout/LineLength:
 Rails/Delegate:
   Exclude:
     - app/controllers/api/concerns/act_as_api_request.rb
+
+Rails/LexicallyScopedActionFilter:
+  Exclude:
+    - app/controllers/api/v1/api_controller.rb
```
santib, line 33: why not following the rules?

app/controllers/api/v1/api_controller.rb
```diff
@@ -1,15 +1,15 @@
 module Api
   module V1
-    class ApiController < ApplicationController
-      include Api::Concerns::ActAsApiRequest
```
santib, line 4: is this concern used somewhere else?

app/controllers/application_controller.rb
```diff
@@ -1,13 +1,13 @@
 class ApplicationController < ActionController::Base
```
santib, line 1: why do we need to keep `ApplicationController`?

config/application.rb
```diff
@@ -33,5 +33,15 @@ class Application < Rails::Application
       g.test_framework :rspec
       g.fixture_replacement :factory_bot, dir: 'spec/factories'
     end
+
+    # Only loads a smaller set of middleware suitable for API only apps.
+    # Middleware like session, flash, cookies can be added back manually.
+    # Skip views, helpers and assets when generating a new resource.
+    config.api_only = true
```
santib, line 40: what happens with ActiveAdmin?

.rubocop.yml
juanmanuelramallo: This is to be able to do this https://github.com/rootstrap/rails_api_base/pull/253/files#diff-e3dabcd50af3f695337100894a51d413R7

which is what we are currently doing in ApplicationController but somehow rubocop is not complaining about

app/controllers/api/v1/api_controller.rb
juanmanuelramallo: Yes in Devise controllers used from the API.

>One consideration: Devise controllers will always
inherit from ApplicationController, hence we still need to use the
ActAsApiRequest concern in Devise controllers. This is because Devise is shared
between the API-only app and the admin panel.

app/controllers/application_controller.rb
juanmanuelramallo: Because ActiveAdmin and Devise (for admin users) use it.

config/application.rb
juanmanuelramallo: It just works, but after we add these https://github.com/rootstrap/rails_api_base/pull/253/files#diff-b1fe55db50c712fef0673345e5b9c0d9R42

.rubocop.yml
santib: what are the conflicting lines?

app/controllers/application_controller.rb
santib: https://github.com/lynndylanhurley/devise_token_auth/blob/7f8a0301191c4a857c94a5fcb2029de04bcf0d00/lib/generators/devise_token_auth/install_generator_helpers.rb#L23 this looks like it should inherit from `ActionController::API` but I guess that'd break ActiveAdmin

config/application.rb
santib: and is the order of those middlewares fine? it doesn't even matter?

juanmanuelramallo: I guess the order matters but maybe not in this case. I tried it locally and everything worked fine.

santib (approved):
I like it! Please make sure it works in Heroku before merging 👀

config/application.rb
juanmanuelramallo: Maybe they matter, and maybe I added the flash deliberately at the end because it uses the cookie (?) haha _sitedigolaverdadtemiento.com_

santib: maybe we should insert them in the rails default order?

.rubocop.yml
juanmanuelramallo: These two https://github.com/rootstrap/rails_api_base/pull/253/files#diff-e3dabcd50af3f695337100894a51d413R7-R8

app/controllers/application_controller.rb
juanmanuelramallo: I'll look into it

config/application.rb
juanmanuelramallo: Ok sounds a good idea, I'll investigate about that ordering

juan-apa:
@juanmanuelramallo  are layouts still usable after this? I think there are some case where they are usefull for API only applications. e.g.: if you are using pagy to handle pagination, you have a `@pagy` variable which contains the pagination data. I've used a layout in the past to show a pagination key in the json responses when the `@pagy` variable is present:

```ruby
# app/controllers/application_controller.rb
layout 'api.v1.'

# app/views/layouts/api.v1.json.jbuilder
json.merge! JSON.parse(yield)
if @pagy
  json.pagination do
    json.total_pages @pagy.pages
    json.current_page @pagy.page
    json.page_size @pagy.vars[:items]
    json.total_count @pagy.count
  end
end

# app/controller/posts_controller.rb
def index
  @pagy, @posts = pagy(Post.all)
end

# GET /api/v1/posts
{
  "posts": [ ... ],
  "pagination": {
    "total_pages": ...,
    "current_page": ...,
    "page_size": ...,
    "total_count": ...
    }
}

juanmanuelramallo:
Good question, short answer is no.

Layouts comes from ActionView and inheriting from ActionController::API removes all of those modules that API applications might not use.

I see your example and although valid my end goal with this change is to move us a bit closer towards using the standard JSON:API.

Do you think we should not move forward with this PR?

juan-apa:
> Good question, short answer is no.
> 
> Layouts comes from ActionView and inheriting from ActionController::API removes all of those modules that API applications might not use.
> 
> I see your example and although valid my end goal with this change is to move us a bit closer towards using the standard JSON:API.
> 
> Do you think we should not move forward with this PR?

it's not that I'm against. If the end goal is to have the responses in JSON:API format, then my previous comment should be implemented differently. Also pagy provides a config to provide the pagination data in the response headers, so using pagy without the pagination info in the response body shouldn't be an issue.

brunvez (commented):
:tumbleweed:

config/application.rb
juanmanuelramallo: This is the default order https://guides.rubyonrails.org/rails_on_rack.html#inspecting-middleware-stack

juanmanuelramallo:
> I like it! Please make sure it works in Heroku before merging

https://using-rails-api-rs.herokuapp.com/
building the test here

horacio (approved):

jpascual1994 (approved):

PR #254, opened 2020-09-22
Enhancement: Add file tracking checks to CI

martinjaimem:
**In this PR:**
 - [x] Added CI check for un-tracked schema.rb
 - [x] Added CI check for un-tracked Gemfile.lock

The reason behind this is that the developer should always commit these files and therefore if by any chance the dev forgets to do so, then the build should fail.

[source](https://thoughtbot.com/blog/keep-autogenerated-files-synced)

To see this in action please check the following builds in another project where I tested this:
 - https://app.circleci.com/pipelines/github/rootstrap/takion-api/53/workflows/8fbcce71-3c86-4ed6-a493-f553089df7e9/jobs/54
 - https://app.circleci.com/pipelines/github/rootstrap/takion-api/51/workflows/a845dc4d-16d5-4853-abb0-3f19162788c6/jobs/52

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,24 @@
-version: 2
+version: 2.1
+
+commands:
+  check_untracked_files:
+    description: Check un-tracked changes using git porcelain command
+    parameters:
+      path:
+        description: Path to check
+        type: string
+        default: "."
+    steps:
+      - run:
+          name: Check un-tracked changes
+          command: |
+            if [[ `git status <<parameters.path>> --porcelain` ]]
+            then
+              echo "ERROR: running the previous command has introduced changes. Hence, Failing the build."
+              git status --porcelain
+              exit 1
+            fi
+
 jobs:
   build:
     environment:
@@ -23,6 +43,10 @@ jobs:
           command: |
             gem install bundler
             bundle install --clean --jobs=4 --retry=3 --path vendor/bundle
+
+      - check_untracked_files:
+          path: "Gemfile.lock"
+
       - save_cache:
           paths:
             - vendor/bundle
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,4 +1,24 @@
...
+    parameters:
+      path:
+        description: Path to check
+        type: string
+        default: "."
+    steps:
+      - run:
+          name: Check un-tracked changes
+          command: |
+            if [[ `git status <<parameters.path>> --porcelain` ]]
```
brunvez, line 15: 🍵

brunvez (approved):
What's this for?

martinjaimem:
@brunvez 

An autogenerated file was committed, but the code that generates it was not.
 - [x] For example, someone edited package.json and ran yarn install locally and then added the changes on yarn.lock but forgot to add the changes on package.json to their commit.
 - [x] The code that generates the file was added, but the autogenerated file was not. As in when someone committed a new migration but forgot to add the updated database schema.

We’ve all made mistakes like these in the past, or missed these aspects during a code review. And that’s ok. In my opinion, if a computer generates a file, then a computer should be in charge of making sure that file is up to date.

I extracted this from: https://thoughtbot.com/blog/keep-autogenerated-files-synced

brunvez:
Nice

sebaherrera07 (approved):
Looks good 👍 .

Do you think this could have any downsides?

.circleci/config.yml
santib, line 15: PORCELAIN?!!!??! 🤯

santib (approved):

martinjaimem:
> Looks good 👍 .
> 
> Do you think this could have any downsides?

Not that I recall but maybe we end up facing some

PR #255, opened 2020-09-22
Enhancement: set CC_TEST_REPORTER_ID as a env key in CircleCI

martinjaimem:
**In this PR:**
 - [x] Set CC_TEST_REPORTER_ID as a env key in CircleCI

**Description:**
As @pablanco suggested, we could be adding this key to the ENV variables of the project.
This will not only improve security, but also when cloning this repository, if the key is not set, then the CI build will fail. So in a way we are enforcing the devs to set their own key

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ With `rake code_analysis` you can run the code analysis tool, you can omit rules
 ## Configuring Code Climate
 1. After adding the project to CC, go to `Repo Settings`
 1. On the `Test Coverage` tab, copy the `Test Reporter ID`
-1. Replace the current value of `CC_TEST_REPORTER_ID` on the `config.yml file (.circleci/config.yml)` with the one you copied from CC
+1. Set the current value of `CC_TEST_REPORTER_ID` in the [circle-ci project env variables](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project)
 
 ## Code Owners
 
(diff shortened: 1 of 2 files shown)


pablanco (approved):
Nice enhancement @martinjaimem !!!

santib (approved):

sebaherrera07 (approved):

README.md
```diff
@@ -100,8 +100,8 @@ With `rake code_analysis` you can run the code analysis tool, you can omit rules
 
 ## Configuring Code Climate
 1. After adding the project to CC, go to `Repo Settings`
-1. On the `Test Coverage` tab, copy the `Test Reporter ID`
-1. Replace the current value of `CC_TEST_REPORTER_ID` on the `config.yml file (.circleci/config.yml)` with the one you copied from CC
+2. On the `Test Coverage` tab, copy the `Test Reporter ID`
+3. Set the current value of `CC_TEST_REPORTER_ID` in the [circle-ci project env variables](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project)
```
santib, line 104: it's better to use
```md
1. x
1. y
1. z
```

because it'll get the correct number automatically

santib: https://guides.github.com/features/mastering-markdown/

fedeagripa (approved):

PR #257, opened 2020-09-23
Enhancement: check if annotate was run in CircleCI

martinjaimem:
I have found PRs that change the annotations on files that are not necessarily changed in that particular PR, in fact, I have made the mistake of uploading pull requests without running the annotate 😞 . So in order to prevent that kind of thing happen, I extended the check of changed files to see if running annotate changed any (which shouldn't).

 FYI: I placed this check after all the other ones as the build is less common that someone forgot to run `annotate` than to have failing tests or have code style errors.

Added Time to the build: `1 second`
Screen Shot 2020-09-23 at 18 01 08

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,3 +86,11 @@ jobs:
             chmod +x ./cc-test-reporter
             ./cc-test-reporter before-build
             ./cc-test-reporter after-build
+
+      - run:
+          name: Check for missing annotations
+          command: RAILS_ENV=test bundle exec annotate
+      
+      # Check if running annotate changed any file
+      - check_untracked_files:
+          path: "app/ spec/"


.circleci/config.yml
```diff
@@ -86,3 +86,11 @@ jobs:
             chmod +x ./cc-test-reporter
             ./cc-test-reporter before-build
             ./cc-test-reporter after-build
+
+      - run:
+          name: Check for missing annotations
+          command: RAILS_ENV=test bundle exec annotate
+      
+      # Check if running annotate changed any file
+      - check_untracked_files:
```
sebaherrera07, line 95: I'm not sure how this yml should be structured because I don't remember having to configure it myself.

But I see that you have 2 `check_untracked_files` items at the same level, and both have different `path`. Is that Ok? Does it run from the top down?

Also another thing is that you put the path to check for annotations to be `.`, and maybe you can set it to `app/models`?

martinjaimem: Yeah, my mistake I didn't clarify, the cost of running the `check_untracked_files` check is almost 0 seconds. As the build follows a top-down strategy, I could have placed it only at the end of the build and check if any files changed, but this way (running it several times (3)) it is more clear what the dev is missing (schema/annotations), because it will say specifically what it was checking.

WDYT?

Annotations are in the models, also specs I think, and in factories. Maybe I am missing some more

fedeagripa (approved):

jpascual1994 (approved):

GuilleLeopold (approved):

santib (approved):

PR #267, opened 2020-10-02
Remove profile endpoint and specs

JuannFerrari:
#### Description:

Hi there!

`user#profile` endpoint does the same as `user#show`, it was implemented four years ago in #44 but gets deprecated after #151 . 

In my opinion, is useless to have it because we already have `user#show.`


juanmanuelramallo (approved):

jpascual1994 (approved):

santib (approved):

apiary.apib
```diff
@@ -29,7 +29,7 @@ Rails Api Base is a boilerplate project for JSON RESTful APIs. It follows the co
 + Response 401
 
 
-## Current user's profile [/api/v1/user/profile]
+## Current user's profile [/api/v1/user/show]
```
santib, line 32:
```suggestion
## Current user's profile [/api/v1/user]
```

show is not part of the path

santib: also 

```
    + Attributes
      + user (Profile Response, required)
```

is wrong below in line 53-54

PR #268, opened 2020-10-05
Scope ActiveAdmin policies

### LINKED ISSUE #212: Use policies from the Admin namespace on ActiveAdmin
As per the discussion on #211 ActiveAdmin should use policies from the `Admin` namespace and not the global namespace as it currently does. So for users, it would use `::Admin::UsersPolicy` instead of `::UsersPolicy`. This would allow us to have an `::Admin::ApplicationPolicy` apart from the current `::ApplicationPolicy`

juan-apa:
#### Board:
* resolves #212 

---
#### Description:
- This PR adds support for namespacing the pundit policies. When creating a policy, now, 2 policies will have to be created: `ModelPolicy` and `ActiveAdmin::ModelPolicy`.
- @santib [suggested in the pundit PR](https://github.com/rootstrap/rails_api_base/pull/211#discussion_r430695258) to use the `admin` namespace for the activeadmin policies. However, if the application will have an admin kind of role, then the namespace `admin` will not be available, because it will be allready taken by the activeadmin policies. I can change this if anyone preferes to keep the `admin` namespace for the activeadmin policies.
---
#### Notes:
* I got "inspired" by these sources:
    * [StackOverflow answer](https://stackoverflow.com/a/49821264)
    * [Active Admin doc](https://activeadmin.info/13-authorization-adapter.html)
    * [`ActiveAdmin::PunditAdapter` source](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/pundit_adapter.rb)


juan-apa:
I know, I have some commits from another PR. I'll fix that before merging :eyes:

README.md
```diff
@@ -38,10 +38,25 @@ This template comes with:
...
+1. Have `docker` and `docker-compose` installed (You can check this by doing `docker -v` and `docker-compose -v`)
+2. Modify the following lines in the `database.yml` file:
+  ``` yaml
+  default: &default
+    adapter: postgresql
+    encoding: unicode
+    pool: 5
+    username: postgres
+    password: postgres
+    host: db
```
brunvez, line 50: If this is the only difference you can set it on an ENV var and do
```suggestion
    host: <%= ENV.fetch('DB_HOST') { 'localhost' }
```

app/models/adapters/active_admin_pundit_adapter.rb
```diff
@@ -0,0 +1,46 @@
+module Adapters
+  class ActiveAdminPunditAdapter < ActiveAdmin::PunditAdapter
```
brunvez, line 2: Tests?

app/models/adapters/active_admin_pundit_adapter.rb
```diff
@@ -0,0 +1,46 @@
...
+        get_policy(subject.class, user, subject)
+      end
+    end
+
+    def scope_collection(collection, _action = Auth::READ)
+      return collection if valid_collection_class?(collection)
+
+      namespaced_policy(collection).new(user, collection).resolve
+    rescue Pundit::NotDefinedError => e
+      raise e unless policy_has_scope?(default_policy_class)
```
brunvez, line 23: This seems like a usual case doesn't it? Maybe we could check before instead of using exceptions as flow control

README.md
```diff
@@ -38,10 +38,25 @@ This template comes with:
 
 ## How to use with docker
 
-1. Have docker and docker-compose installed
-1. build and run the application with `docker-compose up`
-1. Run `docker-compose exec web bash -c "rails db:create db:migrate"`
-1. You can now try your REST services!
+1. Have `docker` and `docker-compose` installed (You can check this by doing `docker -v` and `docker-compose -v`)
```
santib, line 41: these changes don't belong to this PR, right?

santib (commented):
This is looking really good, I'd like to see some unit tests for `app/models/adapters/active_admin_pundit_adapter.rb` with all the possible different cases supported

app/models/adapters/active_admin_pundit_adapter.rb
```diff
@@ -0,0 +1,46 @@
+module Adapters
+  class ActiveAdminPunditAdapter < ActiveAdmin::PunditAdapter
+    def get_policy(subject, user, resource)
+      "ActiveAdmin::#{subject}Policy".constantize.new(user, resource)
+    end
+
+    def retrieve_policy(subject)
```
santib, line 7: I think `retrieve_policy` and `get_policy` can be private, right?

app/models/adapters/active_admin_pundit_adapter.rb
```diff
@@ -0,0 +1,46 @@
...
+      when nil then get_policy(subject, user, resource)
+      when Class then get_policy(subject, user, subject.new)
+      else
+        return Pundit.policy!(user, subject) if active_admin_namespace?(subject)
+
+        get_policy(subject.class, user, subject)
+      end
+    end
+
+    def scope_collection(collection, _action = Auth::READ)
```
santib, line 18: why did you need to add `scope_collection` but you didn't need to add `authorized?` ?

juan-apa:
> This is looking really good, I'd like to see some unit tests for `app/models/adapters/active_admin_pundit_adapter.rb` with all the possible different cases supported

yes, that's what I'm most struggling with tbh.

app/models/adapters/active_admin_pundit_adapter.rb
juan-apa: this here actually raises an exception, because I tried to make it similar to what [ActiveAdmin::PunditAdapter does](https://github.com/activeadmin/activeadmin/blob/93175ed7ff09d113cb77f5e946ee8f054c438775/lib/active_admin/pundit_adapter.rb#L20).

app/models/adapters/active_admin_pundit_adapter.rb
juan-apa: `retrieve_policy` is a [public method in `ActiveAdmin::PunditAdapter`](https://github.com/activeadmin/activeadmin/blob/93175ed7ff09d113cb77f5e946ee8f054c438775/lib/active_admin/pundit_adapter.rb#L32), I'm unsure if making it private will deny access to this function from outside the class when ActiveAdmin tries to call it. `get_policy` definetly can be a private method

santib (approved):
Looks good to me! 🏆

brunvez (approved):
Looking good

Issue #266, opened 2020-10-01
[FEATURE] Add rspec_api_documentation gem


JuannFerrari:
**Describe the solution you'd like**
Add [rspec_api_documentation](https://github.com/zipmark/rspec_api_documentation) will help us to make tests and their docs basically by killing two birds in a shot.

**Additional context**
We must change the structure of requests tests of the API, but I think that the docs enhancement that will give us the gem is convenient to take this trade-off

I could take this issue and integrate the gem❤️

PR #269, opened 2020-10-05
Feature/rspec api documentation integration

JuannFerrari:
#### Board:
* [ISSUE](https://github.com/rootstrap/rails_api_base/issues/266)
---
#### Description:

This PR adds `rspec_api_documentation` gem and all API endpoint acceptance specs to help us to auto-document it. IMO having this gem and the specs as an example will help us to encourage acceptance specs in the projects that use this base as a template, giving good documentation about their APIs.

---
#### Notes:
* `aws-sdk-s3', '~> 1.75` changed in gemfile to make `bundle install` works
* `RspecApiDocumentation` initializer has monkey patching to prevent an issue with Rails 6. (this has been already fixed in https://github.com/zipmark/rspec_api_documentation/issues/456 and I'm waiting for a new release of the gem)
* `apiary.apib` is now generated by `./bin/docs` task


config/initializers/rspec_api_documentation.rb
```diff
@@ -0,0 +1,100 @@
+module RspecApiDocumentation
```
juanmanuelramallo, line 1: Shouldn't this be on the same unless block below?

juanmanuelramallo (commented):
Loved it! 

What do you think of adding something like this https://github.com/juanmanuelramallo/target-mvd/blob/master/spec/support/devise_token_auth_overrides.rb to make the API Blueprint document always use the same access token and client id, just to avoid unnecessary changes in the document (?)

config/initializers/rspec_api_documentation.rb
JuannFerrari: Sure! and it's a nice idea to override that tokens indeed, I will add it.

README.md
```diff
@@ -103,6 +105,15 @@ This template comes with:
 
 http://docs.railsapibase.apiary.io
 
+With [Rspec API Doc Generator](https://github.com/zipmark/rspec_api_documentation) you can generate the docs after writing the acceptance specs.
+
+Just run:
+
+` sh ./bin/docs `
```
juanmanuelramallo, line 112: Do we need the `sh` at the beginning? 

```suggestion
`./bin/docs `
```

README.md
```diff
@@ -103,6 +105,15 @@ This template comes with:
...
+
+And an `apiary.apib` file will be generated.
```
juanmanuelramallo, line 114:
```suggestion
An `apiary.apib` file will be generated at the root directory of the project.
```

config/initializers/rspec_api_documentation.rb
```diff
@@ -0,0 +1,101 @@
+unless Rails.env.production?
+
```
juanmanuelramallo, line 2: nitpick: unneeded empty newline (?)

juanmanuelramallo (approved):
simply B E A U T I F U L

README.md
JuannFerrari: At least for zsh shell, yes 🤔

brunvez (approved):

PR #275, opened 2020-11-05
Updating rubocop

juanmanuelramallo:
This PR updates the version of rubocop to 1.2 which is the latest one as
of today.

Corrections are applied as well.

I need help on how to solve the offense from the spec file spec/requests/users/show_spec.rb:28


juan-apa (approved):

martinjaimem (approved):

jdrosales17 (approved):

spec/requests/api/v1/users/show_spec.rb
```diff
@@ -27,7 +27,7 @@
 
   context 'when an unexpected error occurs' do
     it 'returns status 500 internal server error' do
-      allow(Rails).to receive(:env).and_return 'production'.inquiry
+      allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new('production'))
```
brunvez, line 30: Why?

brunvez (approved):

spec/requests/api/v1/users/show_spec.rb
juanmanuelramallo: Because yes. 

Honestly I'm not sure what this test is for, why should we test our app responds with a 500 error code? I'm more inclined into removing this test altogether. 

Any suggestions?

brunvez: Discussed over slack: the server should not be rescuing from `Exception`

brunvez (approved):

PR #279, opened 2020-12-01
ApplicationCable

santir489:
#### Description:

Rails generates these files by default. (https://guides.rubyonrails.org/action_cable_overview.html#channels)
I think we should keep them, as some gems could expect you to have them.

These files were removed in this PR(https://github.com/rootstrap/rails_api_base/pull/127)

### DIFF
--- a/app/channels/application_cable/channel.rb
+++ b/app/channels/application_cable/channel.rb
@@ -1,0 +1,4 @@
+module ApplicationCable
+  class Channel < ActionCable::Channel::Base
+  end
+end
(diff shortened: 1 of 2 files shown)


martinjaimem (approved):

santib (approved):
❤️

JuanfraM (approved):
💯

juan-apa:
what do you think about adding a comment in `routes.rb` that mounts actioncable to the `/cable` endpoint? Also maybe adding the authentication to connection.rb could be a good idea, that way, if the developer wants to use actioncable, he just has to uncomment the line in the routes file. @juanmanuelramallo left something related to this in [this thread](https://github.com/lynndylanhurley/devise_token_auth/issues/986#issuecomment-357151439) :see_no_evil:

PR #287, opened 2021-01-25
Replace Figaro with Dotenv

santib:


sebaherrera07 (approved):
👏

jpascual1994 (approved):
🚀

.gitignore
```diff
@@ -17,6 +17,7 @@ pickle-email-*.html
 config/initializers/secret_token.rb
 config/secrets.yml
 config/application.yml
```
juan-apa, line 19: I think you should also remove this line from gitignore

juan-apa (approved):

.gitignore
santib: Yes you are right. I left it there in case someone upgraded their app with the base app and suddenly they revealed their credentials. But I guess that's not likely to happen, so I'll remove it 👌

juan-apa: oh, I see, if that's the case, feel free to keep/remove it.

PerezIgnacio (approved):

bootstrap.sh
```diff
@@ -13,11 +13,11 @@ project_name=$1
...
 bundle install
 
-#install node packages
+# install node packages
 yarn
 
-# generate an application.yml from the example file
-cp config/application.yml.example config/application.yml
+# generate a .env from the sample file
+cp .env.sample .env
```
brunvez, line 20: Isn't it `.env.development`?

brunvez (approved):

bootstrap.sh
santib: I usually just use `.env` for development (some credentials might be secret) and I don't want to commit them. And I would use specific env files (like `.env.test`) when I want to always override the values for some reason (like making the tests vars the same for everyone and also for the CI).

https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use

BelenRemedi (approved):

jdrosales17 (approved):

PR #290, opened 2021-03-04
fix: activerecord security alert

mwolman:
#### Description:
Upgrade `rails` to the minor version that upgrades `activerecord` to fix [this](https://github.com/advisories/GHSA-8hc4-xxm3-5ppp) security alert


santib:
I think we should also upgrade in the `package.json`:
- "@rails/actioncable": "^6.0.0-alpha"
- "@rails/activestorage": "^6.0.0-alpha"
- "@rails/ujs": "^6.0.0-alpha"

To match the gem version 6.0.3.5.

Apart from that, what do you think about upgrading also the minor version? So going to 6.1.x ?

Alxzu (approved):

jpascual1994 (approved):

mwolman:
@santib I'm gonna update the `package.json`, thanks!

Related to the `rails 6.1` upgrade, I would like to it in another pull request cause maybe there are [a couple of things](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-0-to-rails-6-1) that I need to change and I prefer to leave this fix isolated. What do you think?

santib:
@mwolman sounds good to me

santib (approved):
Thanks 🙌

PR #292, opened 2021-03-10
Upgrade rails to 6.1.3

mwolman:
### Notes:
While running tests, there's a warning that can't be fixed because it comes from [`devise_token_auth`](https://github.com/lynndylanhurley/devise_token_auth) There's an [open pull request](https://github.com/lynndylanhurley/devise_token_auth/pull/1450) to fix it
```
DEPRECATION WARNING: connection_config is deprecated and will be removed from Rails 6.2 (Use connection_db_config instead)
```


Alxzu (approved):
Looks good to me!

JulianPasquale (approved):

jpascual1994 (approved):
🚀

brunvez (approved):
@flomosq didn't you have these warnings in your Target too?

config/application.rb
```diff
@@ -1,6 +1,19 @@
 require_relative 'boot'
 
-require 'rails/all'
+require 'rails'
+# Pick the frameworks you want:
+require 'action_cable/engine'
+require 'action_controller/railtie'
+require 'action_mailbox/engine'
+require 'action_mailer/railtie'
+# require "action_text/engine"
```
santib, line 9: why do we want to remove action text?

config/application.rb
```diff
@@ -1,6 +1,19 @@
...
+require 'action_view/railtie'
+require 'active_job/railtie'
+require 'active_model/railtie'
+require 'active_record/railtie'
+require 'active_storage/engine'
+require 'rails/test_unit/railtie'
+require 'sprockets/railtie'
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
```
santib, line 19: Below there is a line `config.load_defaults 6.0` it should be `config.load_defaults 6.1` now

santib (commented):
I think there have been changes in some other config files that would be good to get them as well. For example:
`config/initializers/filter_parameter_logging.rb`
Before had something like
```
Rails.application.config.filter_parameters += [:password]
```
And now it has
```
Rails.application.config.filter_parameters += [
  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
]
```

config/application.rb
mwolman: Because of this warning

Screenshot 2021-03-15 at 17 29 53

I discussed it on our `ruby` channel and taking into account that is an API the team suggested going with this solution

PR #294, opened 2021-03-19
Remove Set-cookie header in API response

jeroig:
For each successful response by API (skipping sessions, password, and sign controllers) a 'Set-cookie' value was sent in the header response .

For subsequent requests if we used the value 'Set-cookie' in the header, was not necessary use the tokens for authentication, I mean you have two ways to access by token or cookie in header.

This pull request tries to ensure that you will only have access to the API by token

#### See in how it worked:

> Step 1 - Sign In:

````
curl -i --location --request POST 'http://localhost:3000/api/v1/users/sign_in' \
--header 'Content-Type: application/json' \
--data-raw '{
    "user": {
        "email": "jeroig@rootstrap.com",
        "password": "Test1234"
    }
}'
````
Response: (relevant data)
````
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
access-token: OxKk7Rb1DU5wuhpDnt2QXg
token-type: Bearer
client: 4rq-aphrcVbBqDBBMkh25Q
expiry: 1679170722
uid: jeroig@rootstrap.com
````
- - -

> Step 2
````
curl -i --location --request GET 'http://localhost:3000/api/v1/user' \
--header 'Access-Token: OxKk7Rb1DU5wuhpDnt2QXg' \
--header 'token-type: Bearer' \
--header 'client: 4rq-aphrcVbBqDBBMkh25Q' \
--header 'expiry: 1679170722' \
--header 'uid: jeroig@rootstrap.com'
````
Response: (relevant data)
````
HTTP/1.1 200 OK
....
....
....
Set-Cookie: _session_id={{string_too_long}}; path=/; HttpOnly
....
....
....
````

- - -
> Step 3
 ````
curl --location --request GET 'http://localhost:3000/api/v1/user' \
--header 'Cookie: _session_id={{string_too_long}}'
````
Response: Ok
````
HTTP/1.1 200 OK
{"user":{"id":5,"email":"jeroig@rootstrap.com","name":"Juan1 Roig2","username":"jeroig"}}
````

### DIFF
--- a/app/controllers/api/concerns/act_as_api_request.rb
+++ b/app/controllers/api/concerns/act_as_api_request.rb
@@ -4,7 +4,6 @@ module Api
       extend ActiveSupport::Concern
 
       included do
-        skip_before_action :verify_authenticity_token
         before_action :skip_session_storage
         before_action :check_json_request
       end
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -1,6 +1,7 @@
 module Api
   module V1
     class ApiController < ActionController::API
+      include Api::Concerns::ActAsApiRequest
       include Pundit
       include DeviseTokenAuth::Concerns::SetUserByToken
 
--- a/spec/requests/api/v1/users/show_spec.rb
+++ b/spec/requests/api/v1/users/show_spec.rb
@@ -2,6 +2,12 @@ describe 'GET api/v1/users/:id', type: :request do
   let(:user) { create(:user) }
   subject { get api_v1_user_path, headers: auth_headers, as: :json }
 
+
+  it 'there must not be a Set-Cookie in Header' do
+    subject
+    expect(response.headers.keys).not_to include("Set-Cookie")
+  end
+
   it 'returns success' do
     subject
     expect(response).to have_http_status(:success)
@@ -9,7 +15,6 @@ describe 'GET api/v1/users/:id', type: :request do
 
   it "returns the logged in user's data" do
     subject
-
     expect(json[:user][:id]).to eq(user.id)
     expect(json[:user][:name]).to eq(user.full_name)
   end

### CODE REVIEW

File: app/controllers/api/concerns/act_as_api_request.rb [hunk 1]
```diff
@@ -4,7 +4,6 @@ module ActAsApiRequest
       extend ActiveSupport::Concern
 
       included do
-        skip_before_action :verify_authenticity_token
```
santib, line 7: are you sure we can just remove this line? Wouldn't it cause our frontend apps start getting `ActionController::InvalidAuthenticityToken` when submitting forms?

jeroig: Since my point of view, If you access by API each request is independent of each other, and the uniq way to validate this (isolate) request is sending the correct information in the HEADER that not include 'authenticity_token'.

I think that by removing this line ***we should not have problems from ```ActionController::InvalidAuthenticityToken```***

I imagine two scenarios to use the API:

- You have completely separated the back from the front, like two independent systems: in this case, you have no way to send an authenticity_token from the front form, and no problem with that if you send the correct token in each request

- You have a rails project with the back and the front together: Ok, in this case Yes; remove the line  ```skip_before_action :verify_authenticity_token``` could be a problem if in your front you use a form but; if you have in the same project the back & front together your forms should not (although they could ) call a request with the prefix prefix```/api/v1/{anything}``` .
Conceptually, if you call with the ```api/v1/...``` prefix you must validate only by token.
And if you have this last case, then the forms should call the internal controllers without the API prefix and in them we should put the line ```before_action :verify_authenticity_token```. 
As I said before, it's just a point of view

jeroig: Something to add @santib 

Looking at the [route.rb](https://github.com/rootstrap/rails_api_base/blob/master/config/routes.rb) file, can we ensure that no route from `/admin` through by `/api/v1` ?

santib: yep, you are right

santib (approved):
I just tried it out and worked fine for me, thanks for catching this!

brunvez (approved):
I think this is a good improvement, I have the same concerns as Santi but I think you've addressed them quite well

Issue #264, opened 2020-09-29
Migrate to Github Actions


martinjaimem:
I'd like to discuss the option of changing **CircleCI** into **Github Actions**.

As I see it, most of the projects are migrating to **Github Actions** because of the out-of-the-box integration with Github. It has been in production for some time and we have tried it within different projects with success.

Also, as a Company, we are starting to migrate to Github Actions, so changing the current stack would mean a high-risk change, however, if we introduce this in the base, then we can start facing issues (if any) gradually.

-----
Apart from the already mentioned points, we are facing a security issue which is that our current `CC_TEST_REPORTER_ID` is not being seen from forks of the project, so external builds are failing.
[The solution according to circle-ci](https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/) would mean to re-upload every PR coming from a fork.
This is something already faced in other repos, like `Faraday`. You may see that they are now using Github Actions with the `CC_TEST_REPORTER_ID` and running the CI from forks successfully.

Links:
 - [Faraday repo](https://github.com/lostisland/faraday)
 - [Faraday current CI config](https://github.com/lostisland/faraday/blob/master/.github/workflows/ci.yml)
 - [Faraday discussion about CC_TEST_REPORTER_ID issue](https://github.com/lostisland/faraday/issues/984)

 -----

Finally, I'd like to put some emphasis on the fact that this is not a request to change CircleCI, just want to open this for discussion 😄

PR #297, opened 2021-04-05
update ruby version to 2.7.2

jeroig:
#### Description:

This update attempt to remove this warning

````
warning: parser/current is loading parser/ruby27, which recognizes
warning: 2.7.2-compliant syntax, but you are running 2.7.1.
````

Also was needed to update ffi gem in order to [resolve type 'size_t' (Big Sur - M1)](https://stackoverflow.com/questions/66104935/rails-typeerror-unable-to-resolve-type-size-t-big-sur-m1)

### CODE REVIEW

**Review by santib — APPROVED** [review]

File: .ruby-version [hunk 1]
```diff
@@ -1 +1 @@
-2.7.1
+2.7.2
```
sebaherrera07, line 1: Maybe you need to also update the version in `circleci/config.yml`

santib: I think Seba is right, and also in the `Gemfile`, `Gemfile.lock` (by running bundle), and `Dockerfile`

jeroig: I totally agree !!

jeroig:
If anyone can help me why `circleci` failed I appreciate because I don't have experience over that

jeroig:
I decided to close this PR because circleci fail .

And probably is because this PR come from a fork project (not a branch from itself)
For more info [see](https://github.com/rootstrap/rails_api_base/issues/264)

The same change will be sending since other PR

Issue #278, opened 2020-12-01
Render specific fields on session#create


martinjaimem:
Right now when creating a new session we render all the fields of the user:

```
def render_create_success
    render json: { user: resource_data }
end
```

I think this could lead to some problems when adding a new field to the user that is not intended to be rendered. For example, if the field contains sensitive information then we could have security issues, or if the field contains a big amount of data, then the front end could try to store it and have problems when doing so.

Any thoughts?

PR #301, opened 2021-04-19
Render specific user's fields on sessions#create and registrations#create

mrodriguez90:
ENHANCEMENT

#### REPORTED ISSUE 278:
https://github.com/rootstrap/rails_api_base/issues/278
---
#### Description:
When creating a new session we are rendering all the fields of the user:
def render_create_success
    render json: { user: resource_data }
end
As @martinjaimem mentioned [here](https://github.com/rootstrap/rails_api_base/issues/278) this could lead to some problems when adding a new field to the user that is not intended to be rendered. For example, if the field contains sensitive information.
---
#### Current version
app/controllers/api/v1/sessions_controller.rb and app/controllers/api/v1/registrations_controller.rb are using the following devise_token_auth method to render the resource_data/user:
def render_create_success
    render json: { user: resource_data }
end
---
#### Applied solution
The following views have been created to render specific user's attributes using jbuilder:
app/views/api/v1/registrations/_info.json.builder
app/views/api/v1/registrations/show.json.builder
app/views/api/v1/sessions/_info.json.builder
app/views/api/v1/registrations/show.json.builder

def render_create_success was overwritten to render the created views above.
----
#### Notes
The following specs were changed to accept the new render format in json from the sessions and registrations controller.
spec/acceptance/users_spec.rb
spec/acceptance/sessions_spec.rb


app/views/api/v1/registrations/_info.json.jbuilder
```diff
@@ -0,0 +1,11 @@
+json.ignore_nil!
```
juan-apa, line 1: I believe we are not using the `ignore_nil!` in any other view. Maybe keeping it out would be best for now. Or maybe adding it to the remainder of the views. Also, I believe there is already a partial name `_info` , maybe rendering the create could render that partial as to keep consistent with the `show` action, and keeping it DRY. WDYT?

app/views/api/v1/registrations/_info.json.jbuilder
```diff
@@ -0,0 +1,11 @@
...
+json.id         user.id
+json.email      user.email
+json.name       user.full_name
+json.username   user.username
+json.first_name user.first_name
+json.last_name  user.last_name
+json.uid        user.uid
+json.provider user.provider
```
juan-apa, line 9: missing space here

app/views/api/v1/registrations/_info.json.jbuilder
mrodriguez90: @juan-apa Thank you for your feedback, I agree with you and I've sent a new commit with the suggested changes.

juan-apa (approved):

martinjaimem (approved):

JuannFerrari (approved):

PR #309, opened 2021-05-18
Update Readme to include database configuration with docker-compose

martinjaimem:
Right now when configuring the app with the default **DOCKER** credentials results in an error:
```
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
```

Because we should point to `db` host in the `database.yml`

Also, postgres now requires the default username and password, so right now the result is:

```
fe_sendauth: no password supplied
Couldn't create 'a-price-for_development' database. Please check your configuration.
rails aborted!
PG::ConnectionBad: fe_sendauth: no password supplied
```

The fix for that is to add `postgres` and `postgres` as `username/password` for the DB

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This template comes with:
 ## How to use with docker
 
 1. Have `docker` and `docker-compose` installed (You can check this by doing `docker -v` and `docker-compose -v`)
-2. Modify the following lines in the `database.yml` file:
+1. Modify the following lines in the `database.yml` file:
   ``` yaml
   default: &default
     adapter: postgresql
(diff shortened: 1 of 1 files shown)


brunvez (approved):
I trust you

federicoaldunate (approved):

horacio (approved):

guillermoap (approved):

santib (approved):

juan-apa:
I noticed that this issue happens MOST of the times when running under linux. The difference when installing postgresql on linux and MacOS is that on mac, the password can be empty, whereas in linux it must be passed. That's why you were getting that error. The solution to this can also be setting the password in the `docker-compose.yml` as an env in the `web` service.

PR #326, opened 2021-09-29
Add JSON:API

Yuri-SN:
I was browsing closed PRs and liked the idea of adding the JSON:API specification (#285). I will use it in my project.

Does your project need this functionality?


santib:
@Yuri-SN I like what JSON API proposes in the theory but when it comes to practice, I think it hasn't had a good enough adoption in the community. It might be overkill in some cases and complicate things unnecessarily. For these reasons, I don't think we want to add it to this base app. I'm open to discuss this and change my mind, but previous efforts to introduce JSON API haven't been very convincing..

Yuri-SN:
I agree, this is a controversial issue. I have no compelling case for this specification.

I will close this PR because it is just a copy of the old one (#285).

PR #344, opened 2022-04-12
Upgrade Rails to 6.1.5

santib:
#### Description:
Upgrade Rails to latest 6.x to get security fixes


martinjaimem:
@santib I think you need to run `rails db:drop db:create db:migrate` (or just update the schema)

PR #345, opened 2022-04-13
Add GitHub Actions CI

AgostinaDev:
#### Description:
* Move CI to GitHub Actions.
---
#### Notes:
* Annotate gem was added to the test environment because of the "Check for missing annotations" step.
     * bundle exec annotate won't change anything if it's not enabled & configured for test environments (CI = test environment), so the step wouldn't make any sense.
* The "check_untracked_changes" composite action was created to reuse the `"git status <<path>> porcelain"` command within this repo. It can be moved to a separate independent repo to be used in other repos too.

---
#### Tasks:
- [x] Add CI workflow file
- [x] Add DB configuration file for CI
- [x] Test cache usage & untracked changes detection
- [x] Remove CircleCI integration
- [x] Set GA CI check successful result as mandatory to merge pull request
- [ ] Squash merge when approved
---
#### Preview:

Check the [Actions tab](https://github.com/rootstrap/rails_api_base/actions)


martinjaimem (approved):
🚀

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+          ruby-version: 2.7.2
+      - name: Configure Ruby env
+        run: |
+          echo "gem: --no-document" > ~/.gemrc
+          bundle config set clean 'true'
+      - name: Ruby gem cache
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
```
AgostinaDev, line 42: Something I noticed is that it generated different hashes for files in the push/PR runs ("Ruby gem cache" & "JS package cache" steps). 

I think we need to check this behavior to see if this happens for every PR (doesn't find the cache key in the first run), or if this happened here because it's the first PR that uses GA CI.

[PR - 1st run "Ruby gem cache" step](
https://github.com/rootstrap/rails_api_base/runs/6011073300?check_suite_focus=true), it didn't find the gems cache for the given key
imagen

[PR - 2nd run "Ruby gem cache" step](https://github.com/rootstrap/rails_api_base/runs/6011213086?check_suite_focus=true), it found the cache key because it was generated in the first run
imagen

[Push run - "Ruby gem cache" step](https://github.com/rootstrap/rails_api_base/runs/6010763308?check_suite_focus=true), the cache key is different from the one used in the PR
imagen

AgostinaDev, line 42: Something I noticed is that it generated different hashes for files in the push/PR runs ("Ruby gem cache" & "JS package cache" steps). 

I think we need to check this behavior to see if this happens for every PR (doesn't find the cache key in the first run), or if this happened here because it's the first PR that uses GA CI.

[PR - 1st run "Ruby gem cache" step](
https://github.com/rootstrap/rails_api_base/runs/6011073300?check_suite_focus=true), it didn't find the gems cache for the given key
imagen

[PR - 2nd run "Ruby gem cache" step](https://github.com/rootstrap/rails_api_base/runs/6011213086?check_suite_focus=true), it found the cache key because it was generated in the first run
imagen

[Push run - "Ruby gem cache" step](https://github.com/rootstrap/rails_api_base/runs/6010763308?check_suite_focus=true), the cache key is different from the one used in the PR
imagen

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
+name: CI
```
santib, line 1: If we are adding Github Actions I guess we need to remove CircleCI.

@santiagovidal from a business perspective, did RS migrate from CircleCI to Github Actions already?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          yarn install --pure-lockfile
+      - name: Setup Code Climate test-reporter
+        run: |
+          curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+          chmod +x ./cc-test-reporter
+          ./cc-test-reporter before-build        
+      - name: Setup Database
+        run: |
+          cp .github/test_db/config/database.yml config/database.yml
```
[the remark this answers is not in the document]
AgostinaDev: Mostly because the CircleCI integration currently uses the db configuration file ".circleci/database.yml" which it copies to replace "config/database.yml". We're basically doing the same the previous CI was doing.

I guess we can use config/database.yml.example if it works properly and if it's okay to depend on this example file.

.github/workflows/ci.yml
AgostinaDev: Yes, removing CircleCI is a pending task. I'll need some help with that:

- I don't have access to the settings page in this repo to remove the old CI and configure the new one. Can I be given access to configure it?

- What I can remove CircleCI configuration files from the repo. I guess that'll make the CircleCI check fail (which is a required check), so I guess we'll need to modify the rules before merging.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+      RAILS_ENV: test
+      CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.7.2
```
JulianPasquale, line 33: The `ruby-version` field is not needed with a `.ruby-version` file. https://github.com/ruby/setup-ruby#supported-version-syntax

I'd recommend to only keep the `.ruby-version` file, this way if we need to upgrade the Ruby version (which I also recommend 😜) there is only one file to change

.github/workflows/ci.yml
JulianPasquale: Can you try this?
```yaml
- name: Setup Ruby
   uses: ruby/setup-ruby@v1
   with:
     bundler-cache: true
```
https://github.com/ruby/setup-ruby#caching-bundle-install-automatically

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+          ./cc-test-reporter before-build        
+      - name: Setup Database
+        run: |
+          cp .github/test_db/config/database.yml config/database.yml
+          bundle exec rails db:create db:migrate
+      - name: Check for untracked changes in schema.rb
+        uses: ./.github/actions/check_untracked_changes
+        with:
+          path: "./db/schema.rb"
+      - name: I18n Health
```
JulianPasquale, line 84: Safe to ignore, but I'd love to split these tasks into multiple CI tasks 😍. That'd make this easier to debug and track all the failures

.github/workflows/ci.yml
AgostinaDev: [Change commited here](https://github.com/rootstrap/rails_api_base/pull/345/commits/e590faa831b7880d9440bfb036ee69be3deb4552)

.github/workflows/ci.yml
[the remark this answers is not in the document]
AgostinaDev: Can't we use env variables to configure the database and track `config/database.yml` here like suggested in [this comment](https://github.com/rootstrap/special-olympics-rails-api/pull/6#discussion_r852162701)? @JulianPasquale

JulianPasquale (approved):
🤩

.github/workflows/ci.yml
santiagovidal: Sorry for the late response, here goes:

Yes, we're ok with this change from a business perspective. We're currently using both, and it makes sense to (gradually) migrate and use a single provider.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    services:
+      db:
+        image: postgres:9.4
```
santib, line 11: in a follow up PR we should upgrade PG version

santib (approved):
👏

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
+name: CI
+
+on: [push]
```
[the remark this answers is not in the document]
martinjaimem: I would have thought it did 🤔 

Can you test that @AgostinaDev ?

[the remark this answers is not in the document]
AgostinaDev: @santib You can see in this same PR that it shows the CI / build (push) check. 

If we enable the workflow for PRs too, it'll show the workflow run twice. 

Let me change the code again to take a screen shot and compare.

Wen enabled 'on push':

imagen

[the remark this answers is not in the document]
AgostinaDev: Here's the behavior when the workflow is configured for pushes and PRs, you can see it runs two CI workflows (push and pull_request):

imagen

imagen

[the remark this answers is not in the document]
AgostinaDev: I'm not sure if creating a new PR & not pushing anything after its creation changes this behavior or not. We'll have to check that.

PR #352, opened 2022-06-17
Fix assets

santib:
https://rails-api-base-santib.herokuapp.com/admin/login

The resolutions were causing an issue between dependencies, and the sass-rails gem is required by arctic admin


martinjaimem:
I would argue we should just get rid of arctic admin. Should we open an issue to discuss it?

martinjaimem (approved):

jpascual1994 (approved):

santib:
> I would argue we should just get rid of arctic admin. Should we open an issue to discuss it?

@martinjaimem sure

PR #357, opened 2022-09-06
Upgrade to Rails 7.0.3

esparratacus:
#### Board:
* no ticket assigned for this task
---
#### Description:
Rails upgrade to version 7.
---
#### Notes:
* 'spring-watcher-listen' gem is not yet available for this rails version so it has been removed for the time being
* webpacker removal is out of scope for this release. We will implement importmaps on a following release
---
#### Tasks:
Following the rails guide for[ upgrading the framework](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html)


config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,117 @@
+# Be sure to restart your server when you modify this file.
```
santib, line 1: this file isn't needed, you are already loading the rails 7.0 defaults here https://github.com/rootstrap/rails_api_base/pull/357/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9R25

app/controllers/api/v1/passwords_controller.rb
```diff
@@ -4,6 +4,36 @@ class PasswordsController < DeviseTokenAuth::PasswordsController
       protect_from_forgery with: :exception, unless: :json_request?
       include Api::Concerns::ActAsApiRequest
       skip_before_action :check_json_request, on: :edit
+
+      def edit
```
santib, line 8: can you explain why do we need to override the gem action now? maybe point to some links and documentation

config/environments/production.rb
```diff
@@ -78,6 +78,9 @@
   # the I18n.default_locale when a translation cannot be found).
   config.i18n.fallbacks = true
 
+  # Don't log any deprecations.
+  config.active_support.report_deprecations = false
+
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
```
santib, line 86: at the end of this file there is a huge comment that is no longer auto-generated by rails:
```ruby
  # Inserts middleware to perform automatic connection switching.
  # The `database_selector` hash is used to pass options to the DatabaseSelector
  # middleware. The `delay` is used to determine how long to wait after a write
  # to send a subsequent read to the primary.
  #
  # The `database_resolver` class is used by the middleware to determine which
  # database is appropriate to use based on the time delay.
  #
  # The `database_resolver_context` class is used by the middleware to set
  # timestamps for the last write to the primary. The resolver uses the context
  # class timestamps to determine how long to wait before reading from the
  # replica.
  #
  # By default Rails will store a last write timestamp in the session. The
  # DatabaseSelector middleware is designed as such you can define your own
  # strategy for connection switching and pass that into the middleware through
  # these configuration options.
  # config.active_record.database_selector = { delay: 2.seconds }
  # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
  # config.active_record.database_resolver_context = ActiveRecord::Middleware
  #                                                  ::DatabaseSelector::Resolver::Session
```

config/initializers/content_security_policy.rb
```diff
@@ -1,24 +1,27 @@
...
+#
+#   # Generate session nonces for permitted importmap and inline scripts
+#   config.content_security_policy_nonce_generator = ->(request) { request.session.id.to_s }
+#   config.content_security_policy_nonce_directives = %w(script-src)
+#
+#   # Report violations without enforcing the policy.
+#   # config.content_security_policy_report_only = true
 # end
 
 # Set the nonce only to specific directives
```
santib, line 27: this part is no longer auto-generated by Rails
```ruby
# Set the nonce only to specific directives
# Rails.application.config.content_security_policy_nonce_directives = %w(script-src)
# Report CSP violations to a specified URI
# For further information see the following documentation:
# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
# Rails.application.config.content_security_policy_report_only = true
```

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,8 @@
 # Be sure to restart your server when you modify this file.
 
-# Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+# Configure parameters to be filtered from the log file. Use this to limit dissemination of
+# sensitive information. See the ActiveSupport::ParameterFilter documentation for supported
+# notations and behaviors.
+Rails.application.config.filter_parameters += %i[
+  passw secret token _key crypt salt certificate otp ssn password
```
santib, line 7:
```suggestion
  passw secret token _key crypt salt certificate otp ssn
```

`passw` includes `password`

Gemfile
```diff
@@ -2,16 +2,16 @@ source 'https://rubygems.org'
 ruby '~> 2.7.2'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 6.1.6'
+gem 'rails', '~> 7.0.3'
```
santib, line 5: also make sure to update the rails-related packages in `package.json` to match:
```
    "@rails/actioncable": "^6.1.6",
    "@rails/activestorage": "^6.1.6",
    "@rails/ujs": "^6.1.6",
```

config/environments/production.rb
esparratacus: should I remove these comments then?

santib: I think so, because we didn't add them ourselves it's an autogenerated thing that is no longer autogenerated (for some reason)

config/environments/production.rb
```diff
@@ -81,12 +81,6 @@
   # Send deprecation notices to registered listeners.
```
santib, line 81: line 58, the new log level is `:info` by default https://github.com/rails/rails/pull/39707

.env.test
```diff
@@ -1,2 +1,3 @@
 SERVER_HOST=localhost
-PASSWORD_RESET_URL=http://localhost:3000
+PASSWORD_RESET_URL=/
+
```
santib, line 3:
```suggestion
```

config/initializers/content_security_policy.rb
santib: I think this is the last thing to fix

santib (approved):
Good job 💪

sebaherrera07 (approved):
Looks great!

sebastiancaraballo (approved):

jpascual1994 (approved):

rodrieiz:
Great job!! 👏 
 I did this update some months ago and its looks very similar. I saw a difference on the application and api controllers I had to change the `include Pundit` for `include Pundit::Authorization` but if you don't have problems with the test maybe it's not necessary

PR #359, opened 2022-09-27
Rails 7 webpack removal

esparratacus:
#### Board:
* no ticket
---
#### Description:
Rails 7 has removed the need of webpacker for asset bundling. This PR has active admin be run through sprockets. We removed the webpacker dependencies and install active admin again with sprockets. We will be using esbuild on a future PR.


package.json
```diff
@@ -7,14 +7,8 @@
     "@rails/actioncable": "^7.0.3",
     "@rails/activestorage": "^7.0.3",
     "@rails/ujs": "^7.0.3",
-    "@rails/webpacker": "5.4.2",
```
santib, line 10: you need to run `yarn install` as well

package.json
```diff
@@ -7,14 +7,8 @@
...
     "arctic_admin": "3.2.2",
```
santib, line 10: not sure but maytbe this package is no longer needed when using arctic_admin without webpacker, can you check? https://github.com/cprodhomme/arctic_admin

santib (commented):
There might be some other files to delete. Check https://github.com/rootstrap/rails_api_base/commit/7037a62a3fc9e48b8d7a8cb20bab4696f7849ab8

package.json
santib: same comment for other packages, I guess that without webpacker the `package.json` is no longer in use?

esparratacus: I rather leave it there if we're going to use esbuild in the future

santib: hmmm ok, the thing is that I'm not sure esbuild needs them or if activeadmin+arcticadmin are going to use the traditional asset pipeline

santib (approved):
💪

PR #383, opened 2023-06-29
Replace bullet

santib:
Evaluate replacing Bullet with Rails' `strict_loading` option.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -98,5 +98,5 @@ Rails.application.configure do
   config.active_record.dump_schema_after_migration = false
 
   # Don't raise N+1 errors in production
-  config.active_record.action_on_strict_loading_violation = :raise
+  config.active_record.action_on_strict_loading_violation = :log
 end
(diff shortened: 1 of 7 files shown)


config/environments/production.rb
```diff
@@ -96,4 +96,7 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  # Don't raise N+1 errors in production
+  config.active_record.action_on_strict_loading_violation = :raise
```
JulianPasquale, line 101:
```suggestion
  config.active_record.action_on_strict_loading_violation = :log
```

JulianPasquale (approved):
😍

santib:
@rootstrap/ruby what do you think about this? does someone want to try it out on his/her project before merging it in the base?

Trying it out is mostly checking logs in production for reports and in dev/test see the Dev Experience with the N+1 reports, if it's ok or if we need to tweak certain cases (e.g. when using the rails console)

sebastiancaraballo (approved):

PR #387, opened 2023-07-06
Add feature flag support using Flipper

mrodriguez90:
## Purpose

Add a Feature Flag system to the rails_api_base using Flipper

## Summary

- Adds Flipper
- Adds the [ActiveRecord](https://www.flippercloud.io/docs/adapters/active-record) adapter to store Flipper data
- Adds Flipper-UI to manage feature flags using the interface provided by Flipper.
- Includes the flipper id to the user model, this adds a flipper_id to the user model, this is what you use to enable a feature flag for a given user.
- Adds an initializer for a route constraint that only allows `AdminUser` with a valid session to access the Flipper UI.
- Updates to the Readme

## Context

Ruby community 

## Task

[Notion task](https://www.notion.so/rootstrap/Add-Flipper-to-rails_api_base-7661f5e3a90b4821a570007e1cccabc5?pvs=4)

## Follow-Ups

- Add a cache layer to improve performance, see https://www.flippercloud.io/docs/optimization - [Notion Task](https://www.notion.so/rootstrap/Add-a-cache-layer-to-improve-performance-f411e9e3f78241529b70d197bcff504a?pvs=4)
- Add audit history - [Notion Task
](https://www.notion.so/rootstrap/Add-Audit-history-fa17a205737e43b790d58d95060a9961?pvs=4)
- Configure Flipper for testing - [Notion Task](https://www.notion.so/rootstrap/Configure-Flipper-for-testing-ef52eca64fcc4856801ec8bf13dfc627?pvs=4)

### This is what an `AdminUser` with a valid session will see
Screen Shot 2023-07-06 at 16 06 58
Screen Shot 2023-07-06 at 16 08 39


README.md
```diff
@@ -79,6 +80,9 @@ To illustrate, `bin/rails console` will run the console in the docker container
 - [ExceptionHunter](https://github.com/rootstrap/exception_hunter) for exception tracking
 - [Factory Bot](https://github.com/thoughtbot/factory_bot) for testing data
 - [Faker](https://github.com/stympy/faker) for generating test data
+- [flipper](https://github.com/jnunemaker/flipper) for feature flag support
+- [flipper-active_record](https://github.com/jnunemaker/flipper) for Flipper active_record support
+- [flipper-ui](https://github.com/jnunemaker/flipper) for UI-based feature flag management
```
sebastiancaraballo, line 85: I'm wondering if we should have a `/docs` directory to briefly explain how to access the visual Flipper interface or other relevant items about how / why to use Flipper. Same applies to new features.

app/models/user.rb
```diff
@@ -35,6 +35,7 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :trackable, :validatable
   include DeviseTokenAuth::Concerns::User
+  include Flipper::Identifier
```
sebastiancaraballo, line 38: What is this for? Just asking.

config/initializers/flipper.rb
```diff
@@ -0,0 +1,6 @@
+class CanAccessFlipperUI
+  def self.matches?(request)
+    current_user = request.env['warden'].user
+    current_user.present? && current_user.is_a?(AdminUser)
```
sebastiancaraballo, line 4: I think this should be documented because we are imposing that the user should be of type `AdminUser`.

db/migrate/20230704182448_create_flipper_tables.rb
```diff
@@ -0,0 +1,22 @@
...
+    create_table :flipper_gates do |t|
+      t.string :feature_key, null: false
+      t.string :key, null: false
+      t.string :value
+      t.timestamps null: false
+    end
+    add_index :flipper_gates, %i[feature_key key value], unique: true
+  end
+
+  def self.down
```
sebastiancaraballo, line 18: Should we use reversible migrations? Just asking.

config/routes.rb
```diff
@@ -1,6 +1,9 @@
 Rails.application.routes.draw do
   devise_for :admin_users, ActiveAdmin::Devise.config
   ActiveAdmin.routes(self)
+  constraints CanAccessFlipperUI do
+    mount Flipper::UI.app(Flipper) => '/flipper'
```
sebastiancaraballo, line 5: Do we want this enabled by default? Again just thinking out loud.

sebastiancaraballo (approved):
LGTM. Left some non-blocking comments.

config/routes.rb
```diff
@@ -1,6 +1,9 @@
 Rails.application.routes.draw do
   devise_for :admin_users, ActiveAdmin::Devise.config
   ActiveAdmin.routes(self)
+  constraints CanAccessFlipperUI do
```
JulianPasquale, line 4: I'm wondering if we can use Devise's [authenticate](https://rubydoc.info/github/heartcombo/devise/main/ActionDispatch%2FRouting%2FMapper:authenticate) method here

```ruby
authenticate(:admin_user) do
  mount Flipper::UI.app(Flipper) => '/flipper'
end
```

Also, would be possible to mount this as part of our admin panel? Maybe using `/admin/flipper` route

diego-suarez (approved):
🌝 🏴

Gemfile
```diff
@@ -13,6 +13,9 @@ gem 'devise', '~> 4.7', '>= 4.7.2'
 gem 'devise_token_auth', '~> 1.2', git: 'https://github.com/lynndylanhurley/devise_token_auth'
 gem 'draper', '~> 4.0', '>= 4.0.1'
 gem 'exception_hunter', github: 'rootstrap/exception_hunter'
+gem 'flipper'
+gem 'flipper-active_record'
+gem 'flipper-ui'
```
santib, line 18: please add versions with `~>` to prevent undesired upgrades (when upgrading some other gem).

https://bundler.io/guides/rationale.html
https://thoughtbot.com/blog/a-healthy-bundle
http://web.archive.org/web/20160413161331/blog.zenspider.com/blog/2008/10/rubygems-how-to-preventing-catastrophe.html

config/initializers/flipper.rb
```diff
@@ -0,0 +1,6 @@
+class CanAccessFlipperUI
```
santib, line 1: I think this code should live in `lib/constraints/` or similar (it's said here https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Scoping.html)

The problem is that we'll need to add `lib/constraints` to the autoload paths. Maybe it's just easier to move it to `app/constraints` ?

I don't like an initializer because they have their own lifecycle (they are kinda special), and I don't see the need to have a plain Ruby class defined there.

santib, line 1: I think this class can be renamed to `AdminAccessible`, it's more descriptive to what it's actually doing and should be reused if we add any other admin-only route

config/routes.rb
santib, line 5: What do you think about changing the route to `/feature-flags` or similar so that we make it more friendly to admins? They don't need to know that the tool is called Flipper

config/routes.rb
santib: I like the prefix `/admin` as well. Actually, the constraint could wrap the whole  `admin` namespace, except for the admin sign in/up routes

config/initializers/flipper.rb
santib, line 1: would be good to add tests to this class

app/models/user.rb
santib: I think it's to add the `flipper_id` method that identifies the "actor".

Should we include it in `ApplicationRecord` so that all of our records have the method defined? For example an app could have a model `Product` and we might only want to enable certain feature for some products.. just thinking out loud, would like to hear others' thoughts

blacksam07:
@mrodriguez90 @santib What is the purpose of including `Flipper` in `rails_api_base`? I know `Flipper` is an awesome tool, but not all projects will use this feature and will only have _`thrash`_ gems with useless code

mrodriguez90:
> @mrodriguez90 @santib What is the purpose of including `Flipper` in `rails_api_base`? I know `Flipper` is an awesome tool, but not all projects will use this feature and will only have _`thrash`_ gems with useless code

@blacksam07 I think anyone could benefit from using a tool like this. From my POV, it's an excellent tool and a compelling reason to select the rails_api_base to start any project, particularly for the time you could save by having this tool already set up.

For people who don't require this tool, I think it's much simpler to remove the gem and a few lines of code, rather than investing the effort required to configure it for those who actually want it. The same idea applies to all the gems in the base.

I think it's very unlikely to find a base that covers everything you need and in the way you need it. I think people use the base that most closely matches their needs and then change it as needed.

santib:
> @mrodriguez90 @santib What is the purpose of including `Flipper` in `rails_api_base`? I know `Flipper` is an awesome tool, but not all projects will use this feature and will only have _`thrash`_ gems with useless code

@blacksam07 So first of all, I agree that we need to be thoughtful about what we add to a base app 👍 On the other hand, I don't see "feature flags" (or flipper) as a feature. Actually, for me it's more like a Platform tool, and that's why I think it's a good idea to have it in ALL of our projects. Because for me it shouldn't be considered an optional thing, instead I'd like to promote a **way of working** by promoting continuous integration and delivery flows with the usage of Feature Flags. [There are different kinds of feature flags](https://martinfowler.com/articles/feature-toggles.html), the ones I'm specifically referring to, are the "release toggles".

>they are intended as a tool for Engineers and Site Reliability Engineers to use to de-risk their changes. Feature flags are the shim that gets us to Continuous Delivery with our monorepo and without having to deploy the entire codebase on every change. Feature flags are created to ensure that we can safely rollout our work on our terms.

https://about.gitlab.com/handbook/product-development-flow/feature-flag-lifecycle

blacksam07:
@mrodriguez90 @santib thanks for your replies, I catch the both points, and I know that `Flipper/Feature Flags` is an awesome tool, and we can define as a great tool in our projects, my initial comment was because this is a Community repo not just for out internal projects and we need to be carefully with all new features/tools/config that we include here.

Also, this is a great job 👏

app/models/user.rb
sebastiancaraballo: I think that could be an option, the implementation shouldn't impact or be tightly coupled with the business logic

mrodriguez90: I'm not sure about including this in `ApplicationRecord`. Feature flags are typically required for only a limited number of models, like User, Organization, Product.

README.md
santib, line 85: maybe in this place of the docs we can just say

```
- [Flipper](https://github.com/jnunemaker/flipper) for feature flag support
```

the activerecord and UI things can be implicit I guess?

config/routes.rb
```diff
@@ -1,6 +1,11 @@
 Rails.application.routes.draw do
   devise_for :admin_users, ActiveAdmin::Devise.config
   ActiveAdmin.routes(self)
+  namespace :admin do
+    authenticate(:admin_user) do
```
santib, line 5: maybe we should we add a test to prevent regressions here?

mrodriguez90: @santib I agree, any suggestions on where a test like this should live?

santib (approved):
Maybe we can add a link in the admin panel, so that admins can easily navigate to the FFs page?

config/routes.rb
santib: I think any of a system/feature/request spec would be fine

santib:
Oh I just realized there are [configs for the UI](https://github.com/jnunemaker/flipper/blob/main/lib/flipper/ui/configuration.rb) just in case we want to remove the cloud recommendation or change the texts, adding a link back to the admin or something.

Also, not sure if we need to add some mechanism to auto-create flags disabled by default. For example, if I, as a dev add a flag to the code, when I deploy the code I want the flag to automatically appear in the dashboard being disasbled. Right? Is that currently happening?

config/routes.rb
mrodriguez90: @santib I just realized that rails_api_base doesn't have Capybara to run a system spec. Is Capybara something we would like to add?

santib: hmm let's not add it here in this PR, we can discuss separately if we want to have system specs for our admin or not. Let's go with a request spec for now

spec/requests/api/v1/feature_flags_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      it 'redirects the user to the admin login page' do
+        is_expected.to redirect_to('/admin/login')
+      end
+    end
+  end
+
+  context 'with a user account' do
+    let(:user) { create :user }
+
+    context 'with a valid session' do
```
santib, line 25:
```suggestion
    context 'with an invalid session' do
```

mrodriguez90:
> Oh I just realized there are [configs for the UI](https://github.com/jnunemaker/flipper/blob/main/lib/flipper/ui/configuration.rb) just in case we want to remove the cloud recommendation or change the texts, adding a link back to the admin or something.
> 
> Also, not sure if we need to add some mechanism to auto-create flags disabled by default. For example, if I, as a dev add a flag to the code, when I deploy the code I want the flag to automatically appear in the dashboard being disasbled. Right? Is that currently happening?

@santib I'll add a new task for the UI changes. All FFs are disabled by default

config/initializers/active_admin.rb
```diff
@@ -229,12 +229,12 @@
...
-  #       menu.add label: "My Great Website",
-  #                url: "http://www.mygreatwebsite.com", html_options: { target: :blank }
-  #     end
-  #   end
+  config.namespace :admin do |admin|
+    admin.build_menu :default do |menu|
+      menu.add label: 'Feature Flags',
+               url: '/admin/feature-flags', html_options: { target: :blank }
+    end
+  end
```
mrodriguez90, line 237: this adds the feature flag menu item to the active_admin page

db/migrate/20230704182448_create_flipper_tables.rb
mrodriguez90: this is being auto-generated by the flipper installation process, not sure if we should change it

spec/rails_helper.rb
```diff
@@ -24,6 +24,7 @@
 RSpec.configure do |config|
   config.render_views = true
   config.include Devise::Test::ControllerHelpers, type: :controller
+  config.include Devise::Test::IntegrationHelpers, type: :request
```
mrodriguez90, line 27: adds devise helpers like `sign_in` to requests tests

santib:
> @santib I'll add a new task for the UI changes. All FFs are disabled by default

@mrodriguez90 Just to clarify what I'm referring to, I think that it'd be good that FFs that are added in the codebase, appear in the Flipper UI automatically.

e.g.
1. As a developer I add the feature flag "chatbot-helper"
2. I deploy the feature flag
3. As an admin I should see a new feature flag "chatbot-helper" in the Flipper UI

Is that currently happening? Or do we need to "register" the Feature Flag so that it appears automatically in the UI?

mrodriguez90:
> > @santib I'll add a new task for the UI changes. All FFs are disabled by default
> 
> @mrodriguez90 Just to clarify what I'm referring to, I think that it'd be good that FFs that are added in the codebase, appear in the Flipper UI automatically.
> 
> e.g.
> 
> 1. As a developer I add the feature flag "chatbot-helper"
> 2. I deploy the feature flag
> 3. As an admin I should see a new feature flag "chatbot-helper" in the Flipper UI
> 
> Is that currently happening? Or do we need to "register" the Feature Flag so that it appears automatically in the UI?

@santib Oh I see, if a developer adds a feature flag in the code it won't show up in the UI automatically, someone has to register the flag first.

This is how I imagine the dev experience:
1. As a developer, you register a feature flag locally, and you merge your changes into production.
2. As a developer you register the same feature flag in production, and your turn it on.

Since these feature flags are stored in the database, I don't think there is an easy way to automatically register them from the codebase.

santib:
> > > @santib I'll add a new task for the UI changes. All FFs are disabled by default
> > 
> > 
> > @mrodriguez90 Just to clarify what I'm referring to, I think that it'd be good that FFs that are added in the codebase, appear in the Flipper UI automatically.
> > e.g.
> > 
> > 1. As a developer I add the feature flag "chatbot-helper"
> > 2. I deploy the feature flag
> > 3. As an admin I should see a new feature flag "chatbot-helper" in the Flipper UI
> > 
> > Is that currently happening? Or do we need to "register" the Feature Flag so that it appears automatically in the UI?
> 
> @santib Oh I see, if a developer adds a feature flag in the code it won't show up in the UI automatically, someone has to register the flag first.
> 
> This is how I imagine the dev experience:
> 
> 1. As a developer, you register a feature flag locally, and you merge your changes into production.
> 2. As a developer you register the same feature flag in production, and your turn it on.
> 
> Since these feature flags are stored in the database, I don't think there is an easy way to automatically register them from the codebase.

@mrodriguez90 Maybe we can have them defined in a YML or a Ruby class, read it and create them. Maybe in a rake task that is automatically run on every release we can do something like
```ruby
namespace :feature_flags do
  desc 'Initialize and turn off all the new feature flags so they are available in the UI'
  task initialize: [:environment, :default_tenant] do
    FeatureFlagValue::NAMES.each do |feature_flag|
      next if Flipper.exist?(feature_flag)

      Flipper.disable(feature_flag)
    end
  end
end
```

or for example Gitlab also [defines feature flags in more places](https://gitlab.com/gitlab-org/gitlab/-/blob/master/config/feature_flags/development/ai_assist_flag.yml) than just calls in the codebase. I think that is a good idea not only for having them in the UI but also to keep track of all existing flags BTW.

Maybe we can call leave it for another PR, but I think it'd be good to have a centralized place where we store the flags, and also use that for their initialization. That way we also prevent mistakes (e.g. typos) if a dev has to good to the UI and manually enter the flag name

app/models/user.rb
santib: Can you check if [this](https://github.com/jnunemaker/flipper/blob/a82aa8d93c43725a98219ad3b61160ebf65e9cc4/lib/flipper/railtie.rb#L32-L36) is true? Seems like Flipper already configures `ActiveRecord::Base` to include `Flipper::Identifier`

mrodriguez90: @santib yes, nice catch, it looks like their getting started guide is out of date

PR #389, opened 2023-07-19
Fix reek

santib:
Closes https://github.com/rootstrap/rails_api_base/issues/225

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -125,5 +125,11 @@ detectors:
     enabled: false
 
 exclude_paths:
+  - bin
   - config
+  - coverage
   - db/migrate
+  - log
+  - public
+  - tmp
+  - vendor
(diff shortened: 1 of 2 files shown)


santib:
@JulianPasquale fixed

.reek.yml
```diff
@@ -125,5 +125,12 @@ detectors:
     enabled: false
 
 exclude_paths:
+  - bin
   - config
+  - coverage
   - db/migrate
+  - log
+  - public
+  - spec
```
enzofab91, line 134: Is it ok to exclude always spec files? 🤔

santib: TBH I wasn't sure about it. Usually test code is different from app code and doesn't necessarily follow the same standards and best practices. That said, if we were to have custom Ruby classes to deal with things, it would be good that they are maintainable such as any other class in the app folder..

I'll change this, and if we have further evidence in the future that excluding spec is a good idea we can add it back. Does that sound good to you?

enzofab91 (approved):

jpascual1994 (approved):

sebastiancaraballo (approved):

PR #391, opened 2023-07-21
Adding Rubocop Gems and Removing Rootstrap Rubocop

TimoPeraza:
#### Description:
Based on what we decide in [this document](https://www.notion.so/rootstrap/Standard-vs-RS-Rubocop-382b249ae32b4e7ca3567dfe9bd4c4a5). In this PR, we are removing rootstrap-rubocop, adding more rubocop gems and updating the current ones to help us to improve our code style.

Also in this PR we are fixing each linter that appear, after run the new rubocop configuration.

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


.rubocop.yml
```diff
@@ -1,33 +1,23 @@
-inherit_gem:
-  rubocop-rootstrap:
-    - config/rails_edge.yml
+inherit_from:
+  - .rubocop-rs-config.yml
+  - .rubocop-excludes.yml
```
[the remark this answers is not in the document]
TimoPeraza: Yes, I know it's part of the rs config but I thought could be better if we have all the exclusions in a specific file because it is easier to visualize and you can see all the places where you are excluding a cop. In the other file, there are too many rules and information. But it was just my idea, if you don't like it I can remove it.

spec/requests/api/v1/sessions/create_spec.rb
```diff
@@ -25,20 +27,38 @@
...
       expect(json[:user][:provider]).to eq('email')
+    end
+
+    it 'returns the user first name' do
       expect(json[:user][:first_name]).to eq(user.first_name)
+    end
+
+    it 'returns the user last name' do
       expect(json[:user][:last_name]).to eq(user.last_name)
     end
```
santib, line 56: maybe would be better to keep it as it was and add `:aggregate_failures` ?

https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#aggregate_failures-true-default

spec/requests/api/v1/users/create_spec.rb
```diff
@@ -1,97 +1,117 @@
...
+  it 'creates the user' do
+    expect { endpoint }.to change(User, :count).by(1)
+  end
+
+  it 'returns the user id' do
+    endpoint
+    expect(json[:user][:id]).to eq(user.id)
+  end
 
-    it 'returns the user' do
```
santib, line 40: same

.irbrc
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
sebastiancaraballo, line 1: Is this still needed?

spec/requests/api/v1/settings_spec.rb
```diff
@@ -1,39 +1,46 @@
...
 
     it 'return success' do
-      subject
+      endpoint
       expect(response).to be_successful
     end
 
     it 'returns no need to update' do
-      subject
+      endpoint
```
sebastiancaraballo, line 25: I'd prefer keeping `subject`

spec/requests/api/v1/settings_spec.rb
```diff
@@ -1,39 +1,46 @@
...
       expect(json['must_update']).to be(false)
     end
   end
 
   context 'with incorrect settings' do
-    let!(:setting_version) { create(:setting_version, value: '2.0') }
+    before do
+      create(:setting_version, value: '2.0')
+    end
```
sebastiancaraballo, line 33: Why this change? Just asking

.irbrc
santib: I think they are https://github.com/standardrb/standard/pull/181#issuecomment-635722698

.rubocop-excludes.yml
```diff
@@ -0,0 +1,46 @@
...
+
+RSpec/AnyInstance:
+  Exclude:
+    - spec/requests/api/v1/users/show_spec.rb
+
+RSpec/VerifiedDoubleReference:
+  Exclude:
+    - spec/policies/admin/application_policy_spec.rb
+    - spec/policies/admin/page_policy_spec.rb
+    - spec/policies/application_policy_spec.rb
```
santib, line 46: wouldn't it be better to fix them?

.rubocop-excludes.yml
```diff
@@ -0,0 +1,46 @@
+Lint/AmbiguousBlockAssociation:
```
santib, line 1: I'm a bit worried to have apps start with so many exceptions.. it feels like we are encouraging people to do it lightly. No need to do it right now, but maybe we can tackle it in the future. What do you think?

.rubocop-rs-config.yml
```diff
@@ -0,0 +1,187 @@
+Layout/ClassStructure:
+  Enabled: true
```
santib, line 2: just curious, but why do we need to specify that these rules are enabled? is it because they are disabled by default?

rodrieiz (commented):
nice work! some questions
Is the idea of changing "subject" to "policy," "endpoint," etc., suggested by RuboCop, or is it a separate change? 
Is it possible to move the RuboCop files to a folder, or do they have to be in the project's root directory?

spec/requests/api/v1/settings_spec.rb
PerezIgnacio: Agree, at least for when we are just declaring it. We could also modify the `ImplicitSubject` cop. I do prefer naming the subject for most of other cases, but not sure if it's the best to enforce it always.

.rubocop-excludes.yml
sebastiancaraballo: I think we should have a deeper discussion about the cops and what do they enforce 🤓

spec/requests/api/v1/settings_spec.rb
TimoPeraza: I had to change it because the [NamedSubject](https://www.rubydoc.info/gems/rubocop-rspec/1.6.0/RuboCop/Cop/RSpec/NamedSubject) cop. I don't have any preference about it, I would like to hear the opinion of @santib and @JulianPasquale

spec/requests/api/v1/settings_spec.rb
TimoPeraza: I don't remember the name of the cop now, but there is one where It detects an infraction when you declare `let!` and you don't use it how happens in this example, so you should use `before`

.rubocop-excludes.yml
TimoPeraza: The idea was to fix all of them in another PR, the main goal of this one was just to set the rubocop configuration and fix the ones that appear after setting it

TimoPeraza:
> nice work! some questions Is the idea of changing "subject" to "policy," "endpoint," etc., suggested by RuboCop, or is it a separate change? Is it possible to move the RuboCop files to a folder, or do they have to be in the project's root directory?

About the first point, It is because the [NamedSubject](https://www.rubydoc.info/gems/rubocop-rspec/1.6.0/RuboCop/Cop/RSpec/NamedSubject) cop, which forces you to name your subjects.
About the second one, I think most of the guys agree that the rubocop files should be in the root directory,

spec/requests/api/v1/settings_spec.rb
JulianPasquale: I also feel a bit strange naming the subject. Always felt `subject` is descriptive enough.
That said, I don't have an strong opinion either. I'm actually ok disabling this cop, doesn't sound too critical

.rubocop.yml
```diff
@@ -1,33 +1,129 @@
...
 
+require:
+  - rubocop-factory_bot
+  - rubocop-performance
+  - rubocop-rails
+  - rubocop-rake
+  - rubocop-rspec
+
 AllCops:
+  NewCops: enable
```
JulianPasquale, line 16: Using this config, new cops ("pending" cops) will be enabled when we upgrade the gem. The other option would be to use `disable` and manually enable the new ones.
https://docs.rubocop.org/rubocop/versioning.html#pending-cops

I'm ok keeping it as `enable`. I think that will help us to be aware of the new cops and understand them in a better way so we can decide if we really want it or not. 

I would also like to hear the group opinion here 🙂

JulianPasquale (approved):
🔥

spec/requests/api/v1/settings_spec.rb
santib: I don't have a strong opinion either 😅 whatever the team prefers. Or we can just go with Rubocop default, or check what is the preferred style by Standard.rb 🤷‍♂️

spec/spec_helper.rb
```diff
@@ -30,7 +32,7 @@
   end
   config.include FactoryBot::Syntax::Methods
 
-  config.before :each do
```
sebastiancaraballo, line 33: Is this okay?

sebastiancaraballo (approved):

.rubocop.yml
santib: I think I'm good with the enable option, especially since it's a base app

db/migrate/20161011151353_devise_create_users.rb
```diff
@@ -8,7 +10,7 @@ def change
       ## Recoverable
       t.string   :reset_password_token
       t.datetime :reset_password_sent_at
-      t.boolean  :allow_password_change, default: false
+      t.boolean  :allow_password_change, default: false, null: false
```
mwolman, line 13: Do we want to edit an existing migration? Is it safe? Or would it be better to add a new migration with this change?

spec/requests/api/v1/users/create_spec.rb
```diff
@@ -1,97 +1,93 @@
...
+    expect(response).to have_http_status(:success)
+  end
 
-    it 'creates the user' do
-      expect { subject }.to change(User, :count).by(1)
+  it 'creates the user' do
+    expect { subject }.to change(User, :count).by(1)
+  end
+
+  it 'returns the user' do
```
mwolman, line 34: https://rspec.rubystyle.guide/#expectation-per-example
```suggestion
  it 'returns the user data', :aggregate_failures do
```

spec/requests/api/v1/users/show_spec.rb
```diff
@@ -1,17 +1,24 @@
-describe 'GET api/v1/users/:id', type: :request do
-  let(:user) { create(:user) }
+# frozen_string_literal: true
+
+describe 'GET api/v1/users/:id' do
   subject { get api_v1_user_path, headers: auth_headers, as: :json }
 
+  let(:user) { create(:user) }
+
   it_behaves_like 'there must not be a Set-Cookie in Header'
```
mwolman, line 9:
```suggestion
before { subject }
```

And remove the subjects from each unit test

spec/spec_helper.rb
mwolman: https://rspec.rubystyle.guide/#redundant-beforeeach

spec/requests/api/v1/users/update_spec.rb
```diff
@@ -1,24 +1,31 @@
-describe 'PUT api/v1/user/', type: :request do
+# frozen_string_literal: true
+
+describe 'PUT api/v1/user/' do
+  subject { put api_v1_user_path, params:, headers: auth_headers, as: :json }
+
```
mwolman, line 5: Same comment here

db/migrate/20161011151353_devise_create_users.rb
TimoPeraza: Even if we add a new one, the linter issue will appear the same. Anyway, I think it's safe.

spec/requests/api/v1/users/show_spec.rb
TimoPeraza: I think in this case doesn't work right? because in one of our test we have an allow_instance before the subject

spec/requests/api/v1/settings_spec.rb
```diff
@@ -23,7 +28,9 @@
...
```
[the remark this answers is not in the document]
mwolman: @TimoPeraza this one is missing

spec/requests/api/v1/settings_spec.rb
```diff
@@ -1,15 +1,20 @@
...
       device_version: '1.0'
     }
   end
-  subject { get must_update_api_v1_settings_path, params: }
 
   context 'with correct settings' do
-    let!(:setting_version) { create(:setting_version) }
+    before do
+      create(:setting_version)
+    end
```
mwolman, line 17:
```suggestion
    before { create(:setting_version) }
```

mwolman (approved):

PerezIgnacio (approved):

PR #392, opened 2023-07-27
Rubocop suggestions

mwolman:
#### Description:
Some improvements/suggestions to [this PR](https://github.com/rootstrap/rails_api_base/pull/391) 🤗


.rubocop_todo.yml
```diff
@@ -0,0 +1,33 @@
+# This configuration was generated by
```
santib, line 1: I like the idea of the `.rubocop_todo.yml` but as I understand it, it's meant for things that want to be fixed at some point (that why the "TODO"). So I wonder, is it our case here?

db/.rubocop.yml
```diff
@@ -0,0 +1,8 @@
+inherit_from:
```
santib, line 1: hmm I'm not really convinced about having multiples .rubocop.yml overriding each other.. I feel like just a centralized place is simpler.. not sure though

.rubocop.yml
```diff
@@ -21,3 +20,107 @@ AllCops:
...
+    - spec/**/*
+
+Lint/BinaryOperatorWithIdenticalOperands:
+  Enabled: false
+
+Lint/DeprecatedOpenSSLConstant:
+  Enabled: false
+
+Lint/RaiseException:
+  Enabled: false
```
santib, line 45: note for the future, we should double check these cops. Might be a good idea to turn them on (default)

santib (approved):
Great! left some open ended comments

db/.rubocop.yml
TimoPeraza: I'm not sure that this file is in `db/`, I would move it to the main folder

JulianPasquale: Agree, we should move it to the root.

We can customize the root `.rubocop.yml` file using [Include/Exclude](https://docs.rubocop.org/rubocop/1.55/configuration.html#cop-specific-include-and-exclude) to enable some cops only for specific directories

.rubocop_todo.yml
JulianPasquale: @santib that's a good point.

I like the idea as well, but it kinda force us to consider every violation as "tech debt". I think we should all be aligned on this since it will impact the whole company in some way.

IMO: Every violation should be considered as tech debt. If there are no plans to fix it, does the rule make sense? is it useful? Maybe it just needs to be disabled on the main config file.
I can't think of an scenario where we want to keep the offense and still consider that rule useful.

santib: I agree with what you said. In general terms think there are 2 kinds of exceptions:
1. Tech debt: things that should be fixed at some point
2. Directory level: some directories might need different rules, especially when using DSLs like RSpec or ActiveAdmin

For the rails_api_base I think we should only have the second kind of exceptions (which are not intended to be fixed) so projects start with a clean state

DJA89 (approved):

TimoPeraza (approved):

PR #390, opened 2023-07-20
Add knapsack gem

enzofab91:
#### Board:
* [Ticket #NUMBER_OF_THE_TICKET](link_goes_here)
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
- [x] Add each element in this format
---
#### Risk:
*
---

PR #393, opened 2023-07-31
Add parallel gem

grosendo2006:
#### Board:
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


config/database.yml.example
```diff
@@ -13,7 +13,7 @@ development:
 
 test:
  <<: *default
- database: sample_project_test
+ database: sample_project_test<%= ENV['TEST_ENV_NUMBER'] %>
```
megatux, line 16: why is this change in the db name?

grosendo2006: @megatux this is part of adding this gem, and also we're going to merge this and other PRs into a `release` branch, so when we have the functionality working as expected we'll create a new PR to merge against main. That works for you?

lauperalti (approved):

rodrieiz (approved):

enzofab91 (approved):

Gemfile
```diff
@@ -38,6 +38,7 @@ group :development, :test do
   gem 'annotate', '~> 3.2', '>= 3.0.3'
   gem 'dotenv-rails', '~> 2.7.6'
   gem 'factory_bot_rails', '~> 5.1', '>= 5.1.1'
+  gem 'parallel_tests', '~> 4.2.1'
```
santib, line 41:
```suggestion
  gem 'parallel_tests', '~> 4.2'
```

maybe so we allow minor versions upgrades?

.env.test
```diff
@@ -1,2 +1,3 @@
 SERVER_HOST=localhost
 PASSWORD_RESET_URL=/
+TEST_ENV_NUMBER=3
```
santib, line 3: I don't think this should be set 🤔

santib:
Will te integration with the CI and with knapsack be done as part of this PR?

grosendo2006:
> Will te integration with the CI and with knapsack be done as part of this PR?

@santib this is part of adding this gem, and also we're going to merge this and other PRs (integration with the CI and with knapsack) into a release branch, so when we have the functionality working as expected we'll create a new PR to merge against main. That works for you?

.env.test
grosendo2006: I just added as a part of the first step of adding this gem, I can delete it if you want

santib: but it doesn't make sense to have it here, or why do you want to have it?

santib:
> > Will te integration with the CI and with knapsack be done as part of this PR?
> 
> @santib this is part of adding this gem, and also we're going to merge this and other PRs (integration with the CI and with knapsack) into a release branch, so when we have the functionality working as expected we'll create a new PR to merge against main. That works for you?

Whatever you prefer. What I mean is that we are adding the gem, but the tests aren't running in parallel 
image

santib (approved):

grosendo2006:
> > > Will te integration with the CI and with knapsack be done as part of this PR?
> > 
> > 
> > @santib this is part of adding this gem, and also we're going to merge this and other PRs (integration with the CI and with knapsack) into a release branch, so when we have the functionality working as expected we'll create a new PR to merge against main. That works for you?
> 
> Whatever you prefer. What I mean is that we are adding the gem, but the tests aren't running in parallel image

yes, you're right, but to test it locally you have to run this command `RAILS_ENV=test rake parallel:spec` but this is not the way we're going to run the test, so I just added the gem and after merge this and this [other PR](https://github.com/rootstrap/rails_api_base/pull/390)  we can continue working in order to get both working together

PR #395, opened 2023-08-04
ActiveStorage Current

sebastiancaraballo:
#### Description:
- This PR's aim is to fix a recurring issue when using the Active Storage Disk Service when developing locally and also testing. Example of the issue at the preview.
---
#### Notes:
* https://stackoverflow.com/questions/71860516/activestoragecurrent-host-is-deprecated-how-can-i-use-activestoragecurrent
---
#### Preview:
Captura de pantalla 2023-08-04 a la(s) 11 51 12

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   include Pundit::Authorization
+  include ActiveStorage::SetCurrent
 
   after_action :verify_authorized,
                except: :index,
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -48,3 +48,6 @@ Shoulda::Matchers.configure do |config|
     with.library :rails
   end
 end
+
+ActiveStorage::Current.url_options = { host: ENV.fetch('SERVER_HOST', nil),
+                                       port: ENV.fetch('PORT', 3000) }


szcrubyroostrap (approved):

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   include Pundit::Authorization
+  include ActiveStorage::SetCurrent
```
santib, line 3: This shouldn't be necessary, unless you generate the service url directly (e.g. `user.image.url`) but in general [it's recommended](https://edgeguides.rubyonrails.org/active_storage_overview.html#serving-files) that you use the route helpers to generate an activestorage path (e.g. `url_for(user.image)` or `rails_blob_path(user.image)`)

for that reason I think I'd leave this change out of the base. Thoughts?

sebastiancaraballo: I'm not sure actually. What if the developer wants/needs to use `user.image.url`. Also, isn't `url_for` a view helper only?

sebastiancaraballo: Captura de pantalla 2023-08-04 a la(s) 12 34 41

spec/spec_helper.rb
```diff
@@ -32,5 +32,9 @@
 
   config.before :each do
     ActionMailer::Base.deliveries.clear
+    ActiveStorage::Current.url_options = {
+      host: 'localhost:3000',
+      protocol: 'http'
+    }
```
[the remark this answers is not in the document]
sebastiancaraballo: Nice

spec/spec_helper.rb
```diff
@@ -30,6 +30,11 @@
   end
   config.include FactoryBot::Syntax::Methods
 
+  config.before :all do
+    ActiveStorage::Current.url_options = { host: ENV.fetch('SERVER_HOST', nil),
+                                           port: ENV.fetch('PORT', 3000) }
+  end
```
santib, line 36: I don't think you need the before at all.. just 

```suggestion
  ActiveStorage::Current.url_options = { host: ENV.fetch('SERVER_HOST', nil),
                                         port: ENV.fetch('PORT', 3000) }
```

wdyt?

santib: also please, don't pollute the `spec_helper` with Rails related stuff. Move it either to `rails_helper` or maybe even to `config/environments/test.rb`

spec/spec_helper.rb
```diff
@@ -30,6 +30,11 @@
...
+
   config.before :each do
     ActionMailer::Base.deliveries.clear
```
santib, line 39: 😅

santib (approved):

PR #396, opened 2023-08-04
Delete unnecessary files

santib:
The framework default files are only necessary during an upgrade, and it's not our case

The values being set in one of the files are unnecessary since they match Rails' default values


andresg4 (approved):

sebastiancaraballo (approved):
🚀

config/initializers/new_framework_defaults.rb
```diff
@@ -1,15 +0,0 @@
...
-# This file contains migration options to ease your Rails 5.0 upgrade.
-#
-# Read the Rails 5.0 release notes for more info on each option.
-
-# Make Ruby 2.4 preserve the timezone of the receiver when calling `to_time`.
-# Previous versions had false.
-ActiveSupport.to_time_preserves_timezone = true
-
-# Require `belongs_to` associations by default. Previous versions had false.
-ActiveRecord::Base.belongs_to_required_by_default = true
```
santib, line 12: TODO: I'll need to check if we want to "revert" these changes

santib: checked that these are just the defaults, so no need to set them here

rodrieiz (approved):

PR #404, opened 2023-08-14
Add rake task to register feature flags from yml

mrodriguez90:
#### Board:
* [Ticket](https://www.notion.so/rootstrap/Add-a-rake-task-to-register-feature-flags-from-a-centralized-YML-file-62dfea9091264d6aacdc6a609afa1acd?pvs=4)

#### Description:
Add a rake task to register feature flags from a centralized YML file on every release.

#### Notes:
This PR adds a YML file to register feature flags, this YML will be used on every release to register the feature flags in Flipper/active_record. 
This PR also includes descriptions for feature flags, and a bin script to run the rake task that registers the FFs.

#### Preview:
Screen Shot 2023-08-14 at 13 14 59


bin/release.sh
```diff
@@ -0,0 +1,2 @@
+rake db:migrate
```
JulianPasquale, line 1:
```suggestion
#!/usr/bin/env bash

rake db:migrate
```

config/feature-flags.yml
```diff
@@ -0,0 +1,4 @@
+# feature_flag_test:
```
santib, line 1: ohh why did you remove the "schema" definition? It was cool!

bin/release.sh
santib, line 1: you need to do `chmod +x ./bin/release.sh` git add/commit/push, so that this file is executable

lib/tasks/feature_flags.rake
```diff
@@ -0,0 +1,12 @@
+FEATURES_YML_PATH = 'config/feature-flags.yml'.freeze
```
santib, line 1: `/Users/user/dev/rails_api_base/config/initializers/flipper.rb:1: warning: already initialized constant FEATURES_YML_PATH`

dropping constants like this is a problem, they get defined in the global namespace and are even conflicting between themselves. I'd just use the string directly in this case

bin/release.sh
mrodriguez90: Good call @JulianPasquale

config/feature-flags.yml
mrodriguez90: @santib not sure why 😃 , I've added it back again

bin/release.sh
mrodriguez90: thanks!

lib/tasks/feature_flags.rake
mrodriguez90: good catch

spec/lib/tasks/feature_flags_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rake'
+
+RSpec.describe 'Feature Flags', :rake do
+  before(:all) do
+    task_defined = Rake::Task.task_defined?('feature_flags:initialize')
```
santib, line 5: why wouldn't it be defined?

mrodriguez90: @santib I'm not sure but when I run the test without this `before` block I get the following error:
Screen Shot 2023-08-18 at 14 04 42

Also, I found it to be common practice to do this in these kinds of tests.

spec/lib/tasks/feature_flags_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rake'
+
+RSpec.describe 'Feature Flags', :rake do
```
santib, line 3: to avoid testing rake tasks, I usually take the approach of moving the ruby code to a ruby class. Something like

```rb
# lib/tasks/feature_flags.rake
require 'lib/tasks/feature_flags'

namespace :feature_flags do
  desc 'Regiser new feature flags from config/feature-flags.yml'
  task initialize: :environment do
    Tasks::FeatureFlags.new.init_all
  end
end
```

```rb
# lib/tasks/feature_flags.rb
module Tasks
  class FeatureFlags
    def init_all
      YAML.load_file('config/feature-flags.yml', fallback: {}).each do |key, _options|
        next if Flipper.exist?(key)

        Flipper.add(key)
      end
    end
  end
end
```

and then you just test the ruby class. BTW if doing this approach you can go back to use constants defined under the corresponding Ruby module/class

lib/tasks/feature_flags.rake
```diff
@@ -0,0 +1,10 @@
+namespace :feature_flags do
+  desc 'Regiser new feature flags from config/feature-flags.yml'
```
santib, line 2:
```suggestion
  desc 'Register new feature flags from config/feature-flags.yml'
```

bin/release.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+rake db: migrate
+rake feature_flags: initialize
```
santib, line 4: not related to this PR but I usually recommend running all rails/rake commands prepended with `bundle exec`

https://thoughtbot.com/blog/but-i-dont-want-to-bundle-exec

santib (commented):
Great 👏

santib (approved):

JulianPasquale (approved):

PR #422, opened 2023-08-21
Improve logging

santib:
BEFORE:
image

AFTER:
image

PROS:
- In staging/production envs we get 1 log per line, instead of many

CONS:
- We lose details about views/partials rendering. Shouldn't be a problem in prod since we should use a monitoring tool for this purpose.


config/initializers/lograge.rb
```diff
@@ -0,0 +1,17 @@
...
+  config.lograge.custom_payload do |controller|
+    request = controller.request
+    {
+      current_user_id: controller.try(:current_user).try(:id),
+      current_admin_user_id: controller.try(:current_admin_user).try(:id),
+      host: request.host,
+      path: request.path,
+      query_string: request.query_string,
+      referer: request.referer,
+      user_agent: request.user_agent
```
santib, line 14: do we want this in our logs? or what information do we want?

config/initializers/lograge.rb
```diff
@@ -0,0 +1,16 @@
...
+  config.lograge.enabled = true
+  config.lograge.custom_payload do |controller|
+    request = controller.request
+    {
+      current_user_id: controller.try(:current_user).try(:id),
+      host: request.host,
+      path: request.path,
+      query_string: request.query_string,
+      referer: request.referer,
+      user_agent: request.user_agent
```
santib, line 13: do we want this in our logs? or what information do we want?

Gemfile
```diff
@@ -19,6 +19,7 @@ gem 'flipper', '~> 0.28.3'
 gem 'flipper-active_record', '~> 0.28.0'
 gem 'flipper-ui', '~> 0.28.0'
 gem 'jbuilder', '~> 2.10'
+gem 'lograge', '~> 0.13.0'
```
rodrieiz, line 22: Hadn't you recommended us to specify the version with 2 digits and allow minor updates?

config/initializers/lograge.rb
rodrieiz: Is possible to add the time info?

rodrieiz (commented):
Just a question, Do we need to remove code from the previous 'prints logs'?

config/initializers/lograge.rb
santib: good question. In the lograge docs it says
>It doesn't include any timestamp by default, instead it assumes you use a proper log formatter instead.

And this is what I see for example in AWS Cloudwatch:
image

Gemfile
santib: good point. From https://semver.org/ it says:
>MINOR version when you add functionality in a backward compatible manner

so it should be safe to do that yes. I'll change it

rodrieiz (approved):

config/initializers/lograge.rb
JulianPasquale: Few questions here:
- Can que add the IP address? Might be useful to detect bots attacks or similar
- Would query_string include all the request params? I mean, from url and from a form/post request body?

JulianPasquale (approved):

config/initializers/lograge.rb
santib: - IIRC IP addresses are considered to be PII which goes against GDPR. I'd like this base app to be GDPR-compliant by default
- query_string only have the query params in the URL

enzofab91 (approved):

PR #444, opened 2023-09-08
Update ruby to 3.2.2 and postgres image to 15

guillermoap:
- Update ruby version to 3.2.2 both in docker and in .ruby-version file.
- Update postgres image version in docker-compose to 15


Gemfile
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
-ruby '~> 3.1.2'
+ruby '~> 3.2.2'
```
santib, line 4: I think some gems will break. For example [rspec_api_documentation](https://github.com/zipmark/rspec_api_documentation/pull/507)

guillermoap: we should probably remove that gem as well since it's pretty outdated.

santib: we need to find a replacement, not just removing it

guillermoap: Will create a new pr removing that and exception_hunter, updating pagy and some other stuff I found.

guillermoap: rswag seems like the closest thing to it

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.1.2
+FROM ruby:3.2.2
```
JulianPasquale, line 1: Can we update the `.dockerignore` file here as well? 🙏

santib: maybe this should be done in a separate PR. I don't think this one will be merged for quite some time

santib:
When running tests we get a warning
`To use retry middleware with Faraday v2.0+, install `faraday-retry` gem`

santib (approved):
🤩

sebastiancaraballo (approved):

jpascual1994 (approved):

PR #453, opened 2023-09-12
Update .dockerignore file

JulianPasquale:
#### Description:

Update `.dockerignore` file to exclude all the unneeded files.
Ignoring all these files will improve docker builds times and generate lightweight images that can be deliver easily.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,31 @@
-node_modules
+# See https://docs.docker.com/engine/reference/builder/#dockerignore-file for more about ignoring files.
+
+# Ignore bundler config.
+/.bundle
+
+# Ignore all default key files
+config/master.key
+config/credentials/*.key
+
+# Ignore all logfiles and tempfiles.
+/log/*
+/tmp/*
+!/log/.keep
+!/tmp/.keep
+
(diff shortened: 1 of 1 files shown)


mrodriguez90 (approved):
😎

jpascual1994 (approved):

.dockerignore
```diff
@@ -1 +1,31 @@
-node_modules
+# See https://docs.docker.com/engine/reference/builder/#dockerignore-file for more about ignoring files.
```
santib, line 1: should we ignore `.env` as well?

santib (approved):

.dockerignore
santib: I wasn't thinking on the image size, but that it could be a security problem to have the .env bundled, similar to why

```ruby
# Ignore all default key files
config/master.key
config/credentials/*.key
```

are ignored I guess

JulianPasquale: I tried to ignore all the `.env` files except for `.env.test`. Let me know what you think now 🙂

santib: great 💪

santib (approved):

PR #457, opened 2023-09-15
Add Strong Migration gem

brunoparma88:
#### Board:
* [NOTION TICKET](https://www.notion.so/rootstrap/3950194b61e342628f6f56ce11200334?v=9f47a8bf83d34841819b3cf549859d94&p=2ef908ece8a443a9b4004cbefd0f9ba2&pm=s)
---
#### Description:
Add [Strong Migration](https://github.com/ankane/strong_migrations) gem
---
#### Notes:
This gem has its own configuration file. Currently, this gem appears disabled, with a future date set as the start date for the gem. To enable the gem, you must change that date to the current date. If an earlier date is set, this gem will analyze risks in already created migrations
---
#### Tasks:

---
#### Risk:
*
---


config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# To mark migrations as safe that were created before installing this gem
+StrongMigrations.start_after = 20300101000000 # Disabled By default #2030/01/01
```
santib, line 4: ?? with this you are disabling strong migrations till 2030, right? why would you do that?

config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
santib, line 1: why does this file differ so much from the one that comes by default? https://github.com/ankane/strong_migrations/blob/master/lib/generators/strong_migrations/templates/initializer.rb.tt

config/initializers/strong_migrations.rb
brunoparma88: That configuration is there because we still don't know if we are going to leave it enabled initially or not. In the event that it is enabled, that should be removed

santib: with that config you are completely disabling the gem right? so what's the point of adding it in the first place?

santib: @brunoparma88 after the conversation we had I understand your point but I think we should have strong migrations enabled by default after the last existing migration

db/migrate/20161027190856_create_delayed_jobs.rb
```diff
@@ -2,7 +2,7 @@
 
 class CreateDelayedJobs < ActiveRecord::Migration[5.0]
   def self.up
-    create_table :delayed_jobs, force: true do |table|
+    create_table :delayed_jobs do |table|
```
santib, line 5: same as above

config/initializers/strong_migrations.rb
santib: @rootstrap/ruby thoughts?

sebastiancaraballo: Agree with @santib. It could be nice to have it enabled for existing migrations though 🤔

mrodriguez90: +1 to Santi and Seba's comments

santib (approved):
Great!

Gemfile
```diff
@@ -29,6 +29,7 @@ gem 'rack-cors', '~> 2.0'
 gem 'sass-rails', '~> 6.0.0'
 gem 'sendgrid', '~> 1.2.4'
 gem 'sprockets', '~> 4.2.1'
+gem 'strong_migrations', '~> 1.6', '>= 1.6.3'
```
jpascual1994, line 32: Shouldn't it be available only for development?

JulianPasquale: Yeah, otherwise we should update the Readme description so it doesn't specify "in development"

santib: I think the gem is needed in all envs by design. 

The "in development" I think comes from the gem's README that says:
>Catch unsafe migrations in development

I think the point is that if you write an unsafe migration you'll catch it during development because you won't be able to even run it locally. Not sure what happens if you try to push to prod, I guess it breaks the build

brunoparma88: It not only reads migrations but also inspects the schema. It makes sense that it's not limited solely to the development environment, as it could also detect errors if a production push is done.

PR #464, opened 2023-09-19
Knapsack ci workflow updates

rodrieiz:
#### Description:
Update Knapsack workflow

#### Notes:
If the branch exists or the PR is already created the workflow will fail.
In your repository settings -> actions -> general ->  Workflow permissions we need to allow read / write permissions and mark the option allow Github action to create Pr, other case the workflow will fail.

---
#### Tasks:
---
#### Risk:
*
---
#### Preview:
The PR looks like this
image

### DIFF
--- a/.github/workflows/update_knapsack_report.yml
+++ b/.github/workflows/update_knapsack_report.yml
@@ -3,7 +3,7 @@ name: Update Knapsack Report
 # This workflow should be scheduled at certain intervals
 on:
   schedule:
-    - cron: '0 5 31 2 *'
+    - cron: '*/15 * * * *'
 # The above cron does not run. Replace with the wanted periodicity.
 # For example, the following configuration would schedule it every 3 months
 #    - cron: '0 0 1 */3 *'
@@ -50,7 +50,7 @@ jobs:
     - name: Push changes
       run: git push origin HEAD:update-knapsack-report
     - name: Initialize Pull Request
-      uses: rootstrap/create-pull-request@v3
+      uses: rootstrap/create-pull-request@v5
       with:
         pull_request_token: ${{ secrets.GITHUB_TOKEN }}
         head: update-knapsack-report
--- a/.github/workflows/update_knapsack_report.yml
+++ b/.github/workflows/update_knapsack_report.yml
@@ -29,7 +29,6 @@ jobs:
 
     env:
       RAILS_ENV: test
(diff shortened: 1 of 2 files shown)


.github/workflows/update_knapsack_report.yml
```diff
@@ -29,14 +29,14 @@ jobs:
 
     env:
       RAILS_ENV: test
+      github_token: ${{ secrets.GITHUB_TOKEN }}
```
santib, line 32:
```suggestion
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/update_knapsack_report.yml
```diff
@@ -29,14 +29,14 @@ jobs:
...
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Ruby 3.1
-      uses: actions/setup-ruby@v1
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 3.1
```
santib, line 38: why? we want to run the same version of ruby as the app because it's gonna run the tests, right?

.github/workflows/update_knapsack_report.yml
santib: where is this ENV VAR used? (not the secrets)

.github/workflows/update_knapsack_report.yml
rodrieiz: Because this was throwing a warning "invalid key". Yes we want this, but this detect the app version automatically
image

.github/workflows/update_knapsack_report.yml
rodrieiz: Thanks. I'll remove

santib (approved):
👏

santib:
@rodrieiz any idea why did this fail? https://github.com/rootstrap/rails_api_base/actions/runs/6239744398/job/16938322883

rodrieiz:
> @rodrieiz any idea why did this fail? https://github.com/rootstrap/rails_api_base/actions/runs/6239744398/job/16938322883

Because in the previous 'set up ruby' step, it wasn't installing the gems, that is why the 'rails db:create' command was failing. But that's an old run, it shouldn't happen anymore

PR #465, opened 2023-09-19
Update docker env var during bootstrap

chrisjlebron:
#### Board:
* N/A
---
#### Description:

When using Docker, the setup script was failing / not working as expected. The `running_with_docker?` method checks the env var `DOCKER_ENABLED=true`, but the default is `false`. This adds a rewrite of that value if bootstrapping with Docker. I also added a utility method for the bootstrap script since this uses `sed` again, necessitating the OS check.

---
#### Notes:
* I'm new here, so I'm not sure if we want to load dotenv in all the bin scripts that reference `running_with_docker?`, but it seems like it might be worthwhile, if anyone runs those scripts in isolation.
---
#### Tasks:
N/A

---
#### Risk:
N/A

---
#### Preview:
N/A

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -58,16 +58,24 @@ fi
 # generate a .env from the sample file
 cp .env.sample .env
 
-# update project name in database.yml
 # sed command implementation is different for GNU and macOS
-if [[ $OSTYPE == 'darwin'* ]]; then
-  sed -i '' "s/rails_api_base/${project_name}/g" config/database.yml
-else
-  sed -i "s/rails_api_base/${project_name}/g" config/database.yml
-fi
+sed_i() {
+  if [[ $OSTYPE == 'darwin'* ]]; then
+    sed -i '' $@
+  else
+    sed -i $@
+  fi
+}
+
+# update project name in database.yml
+sed_i "s/rails_api_base/${project_name}/g" config/database.yml
 
 # spin up docker services if flag is specified for the setup to take place inside the containers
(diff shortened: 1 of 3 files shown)


santib:
>I'm new here, so I'm not sure if we want to load dotenv in all the bin scripts that reference running_with_docker?, but it seems like it might be worthwhile, if anyone runs those scripts in isolation.

@chrisjlebron good catch, maybe we should `require dotenv/load` in `bin/helpers.rb` ? so we don't need to require it in N different places

bootstrap.sh
```diff
@@ -58,16 +58,24 @@ fi
...
 cp .env.sample .env
 
-# update project name in database.yml
 # sed command implementation is different for GNU and macOS
-if [[ $OSTYPE == 'darwin'* ]]; then
-  sed -i '' "s/rails_api_base/${project_name}/g" config/database.yml
-else
-  sed -i "s/rails_api_base/${project_name}/g" config/database.yml
-fi
+sed_i() {
```
guillermoap, line 62: Very nice

guillermoap (approved):

JulianPasquale (approved):
LGTM

chrisjlebron:
> @chrisjlebron good catch, maybe we should `require dotenv/load` in `bin/helpers.rb` ? so we don't need to require it in N different places

Moved that, thanks for the suggestion

santib:
Thanks @chrisjlebron 🙌

PR #492, opened 2023-10-13
Rspec retry pr comment

rodrieiz:
#### Board:
---
#### Description:

In this PR, we add the 'rspec-retry' gem and configure it to retry tests twice.
In case of a failure on the first attempt, and a subsequent pass on the second, a comment will be added with information about the problematic test. (If the test fails always or only the first attempt the comment won’t be added).

We modified the CI to run only on pull requests or direct pushes to the main branch

#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


.github/workflows/ci.yml
```diff
@@ -100,6 +117,12 @@ jobs:
     runs-on: ubuntu-latest
     needs: tests
     steps:
+      - name: Set ENV for codeclimate (pull_request)
+        if: github.event_name == 'pull_request'
+        run: |
+          git fetch --no-tags --prune --depth=1 origin +refs/heads/$GITHUB_HEAD_REF:refs/remotes/origin/$GITHUB_HEAD_REF
+          echo "GIT_BRANCH=$GITHUB_HEAD_REF" >> $GITHUB_ENV
+          echo "GIT_COMMIT_SHA=$(git rev-parse origin/$GITHUB_HEAD_REF)" >> $GITHUB_ENV
```
santib, line 125:
```suggestion
      - name: Checkout code
        uses: actions/checkout@v3
      - name: Set ENV for codeclimate (pull_request)
        if: github.event_name == 'pull_request'
        run: |
          git fetch --no-tags --prune --depth=1 origin +refs/heads/$GITHUB_HEAD_REF:refs/remotes/origin/$GITHUB_HEAD_REF
          echo "GIT_BRANCH=$GITHUB_HEAD_REF" >> $GITHUB_ENV
          echo "GIT_COMMIT_SHA=$(git rev-parse origin/$GITHUB_HEAD_REF)" >> $GITHUB_ENV
```

.reek.yml
```diff
@@ -108,7 +108,8 @@ detectors:
     accept: []
   UncommunicativeVariableName:
     enabled: true
-    exclude: []
+    exclude:
+      - ./spec/support/retry/pull_request_comment.rb
```
santib, line 112: maybe we can fix this instead of ignoring it?

Gemfile
```diff
@@ -42,11 +42,13 @@ group :development, :test do
   gem 'dotenv-rails', '~> 2.8.1'
   gem 'factory_bot_rails', '~> 6.2'
   gem 'knapsack', '~> 4.0'
+  gem 'octokit', '~> 7.1'
   gem 'parallel_tests', '~> 4.3'
   gem 'pry-byebug', '~> 3.9', platform: :mri
   gem 'pry-rails', '~> 0.3.9'
   gem 'rspec_api_documentation', '~> 6.1.0'
   gem 'rspec-rails', '~> 6.0'
+  gem 'rspec-retry', github: 'rootstrap/rspec-retry', branch: 'add-intermittent-callback'
```
santib, line 51: I think these 2 gems are only needed in test env, not in dev right?

santib (approved):
Beautiful!

.github/workflows/ci.yml
```diff
@@ -54,6 +64,13 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+      - name: Set ENV for codeclimate (pull_request)
+        if: github.event_name == 'pull_request'
+        run: |
+          git fetch --no-tags --prune --depth=1 origin +refs/heads/$GITHUB_HEAD_REF:refs/remotes/origin/$GITHUB_HEAD_REF
+          echo "GIT_BRANCH=$GITHUB_HEAD_REF" >> $GITHUB_ENV
+          echo "GIT_COMMIT_SHA=$(git rev-parse origin/$GITHUB_HEAD_REF)" >> $GITHUB_ENV
+          echo "PULL_REQUEST_NUMBER=${{ github.event.pull_request.number }}" >> $GITHUB_ENV
```
enzofab91, line 73: this echos are needed?

enzofab91 (approved):
🎩

spec/support/retry/pull_request_comment.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Retry
+  class PullRequestComment
+    attr_reader :repo, :pull_request_number, :client
+
+    def initialize
+      @repo = ENV.fetch('REPO', nil)
```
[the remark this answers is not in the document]
jpascual1994: It's used on the comment method :)

.github/workflows/ci.yml
```diff
@@ -1,10 +1,20 @@
...
+    branches: [main]
+
+permissions:
+  pull-requests: write
+  contents: write
 
 env:
   RAILS_ENV: test
   CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+  CI: 'true'
```
santib, line 15:
```suggestion
  CI: true
```

.github/workflows/ci.yml
santib: I think so 😄

.rubocop.yml
```diff
@@ -79,6 +79,10 @@ Rails/FilePath:
 Rails/SaveBang:
   Enabled: true
 
+RescuedExceptionsVariableName:
+  Exclude:
+    - "spec/support/retry/pull_request_comment.rb"
```
santib, line 84: maybe we should disable this cop at all? otherwise it will always conflict with reek right?

or maybe we can accept `e` ? https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Variable-Name.md

spec/support/retry/pull_request_comment.rb
```diff
@@ -0,0 +1,21 @@
...
+      @pull_request_number = ENV.fetch('PULL_REQUEST_NUMBER', nil)
+      @github_token = ENV.fetch('GITHUB_TOKEN', nil)
+      @client = Octokit::Client.new(access_token: @github_token)
+    end
+
+    def comment(message)
+      options = { event: 'COMMENT', body: message }
+      client.create_pull_request_review(repo, pull_request_number, options)
+    rescue StandardError => error
+      p error.message
```
santib, line 18:
```suggestion
      puts error.message
```

so it also adds a newline

.github/workflows/ci.yml
rodrieiz: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

.rubocop.yml
rodrieiz: We discussed this earlier today. If I name it "ex," RuboCop also asks me to change it to "e."
`Naming/RescuedExceptionsVariableName: Use 'e' instead of 'ex.'`

If I add to allow that 'e' is a valid variable name, it could potentially be used for anything, not just in this context. I believe this is the best option, and if it repeats, we may need to disable this cop.

.github/workflows/ci.yml
rodrieiz: I tested in a fork and for this job this vars are not necessary

.rubocop.yml
santib: but why wouldn't you disable it at all? It will always cause problems, right? every time we catch an exception

jpascual1994 (approved):

.github/workflows/ci.yml
```diff
@@ -1,10 +1,20 @@
 name: CI
 
-on: [push]
+on:
+  pull_request:
+  push:
+    branches: [main]
```
santib, line 6:
```suggestion
on:
  workflow_dispatch:
  pull_request:
  push:
    branches: [main]
```

unrelated, but can we also add this, so the CI can be manually triggered?

lauperalti (approved):
💪

.github/workflows/ci.yml
rodrieiz: done

.rubocop.yml
rodrieiz: Because there was only one case so far, but yes is better to disable it to avoid future conflicts

spec/support/retry/pull_request_comment.rb
```diff
@@ -0,0 +1,21 @@
...
+      @pull_request_number = ENV.fetch('PULL_REQUEST_NUMBER', nil)
+      @github_token = ENV.fetch('GITHUB_TOKEN', nil)
+      @client = Octokit::Client.new(access_token: @github_token)
+    end
+
+    def comment(message)
+      options = { event: 'COMMENT', body: message }
+      client.create_pull_request_review(repo, pull_request_number, options)
+    rescue StandardError => error
+      puts error.message
```
santib, line 18:
```suggestion
    def comment(message)
      return if pull_request_number.blank?

      options = { event: 'COMMENT', body: message }
      client.create_pull_request_review(repo, pull_request_number, options)
    rescue StandardError => error
      puts error.message
```

so we don't try to do it on local env or main branch

rodrieiz: yes thanks, I was thinking to do something like this

PR #498, opened 2023-10-19
Enable daily version updates

sebastiancaraballo:
#### Board:
* N/A
---
#### Description:
- I noticed that Dependabot tried to open new pull requests to update gem versions but the current limit is 5 weekly PRs.
- This PR aim is to increase that limit to up to 10 PRs every day.
---
#### Notes:
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval
---
#### Preview:
Captura de pantalla 2023-10-19 a la(s) 13 07 06

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,5 @@ updates:
   - package-ecosystem: "bundler"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
+    open-pull-requests-limit: 10


.github/dependabot.yml
```diff
@@ -3,4 +3,5 @@ updates:
   - package-ecosystem: "bundler"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
```
enzofab91, line 6: What about leaving it as it is weekly, but with more limit?

enzofab91 (approved):
Good catch 👏

.github/dependabot.yml
```diff
@@ -3,4 +3,5 @@ updates:
...
+    open-pull-requests-limit: 10
```
santib, line 7: I'm not really sold on this change for 2 reasons:
1. I don't want to be flooded by upgrade PRs in the `rails_api_base`. If we want to make faster progress nothing impedes us to manually go and update stuff.
2. I don't want this change to reach actual projects that are based on the `rails_api_base`

I think 5 PRs/week is a good tradeoff

sebastiancaraballo: I agree, specially on point 2.

PR #499, opened 2023-10-19
Hotfix Ransack 4.0.0

sebastiancaraballo:
#### Board:
---
#### Description:
This [PR](https://github.com/rootstrap/rails_api_base/pull/483) introduced new version of Ransack gem (4.0.0). This introduced some breaking changes and prevents the use of current ActiveAdmin sections.

This [article](https://rogodev.com/ransack-40-allowlist-changes) explains the reason and how to fix the issue.

Let me know if this is the best possible solution. 😄

#### Risk:
* High
---
#### Preview:
Captura de pantalla 2023-10-19 a la(s) 13 52 34

### DIFF
--- a/app/models/admin_user.rb
+++ b/app/models/admin_user.rb
@@ -29,4 +29,7 @@ class AdminUser < ApplicationRecord
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable,
          :recoverable, :rememberable, :trackable, :validatable
+
+  RANSACK_ATTRIBUTES = %w[id email sign_in_count current_sign_in_at last_sign_in_at
+                          current_sign_in_ip last_sign_in_ip created_at updated_at].freeze
 end
--- a/.reek.yml
+++ b/.reek.yml
@@ -10,7 +10,9 @@ detectors:
     exclude: []
   ControlParameter:
     enabled: true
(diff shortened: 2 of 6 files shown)


app/models/admin_user.rb
```diff
@@ -25,8 +25,21 @@
...
+    id
+    email
+    sign_in_count
+    current_sign_in_at
+    last_sign_in_at
+    current_sign_in_ip
+    last_sign_in_ip
+    created_at
+    updated_at
+  ].freeze
```
santib, line 44:
```suggestion
  RANSACK_ATTRIBUTES = %w[id email sign_in_count current_sign_in_at last_sign_in_at
                          current_sign_in_ip last_sign_in_ip created_at updated_at].freeze
```

app/models/admin_user.rb
```diff
@@ -25,8 +25,21 @@
 #
 
 class AdminUser < ApplicationRecord
+  include Ransackable
```
santib, line 28: Maybe we could just `include Ransackable` in `ApplicationRecord` so we don't need to remember to do it on every single model?

santib (approved):
Good catch. It's a pity that our tests didn't catch it. I guess it's time to add some feature specs for our Admin dashboard?

PR #504, opened 2023-10-24
Add missing ruby versions to CI file

guillermoap:
On https://github.com/rootstrap/rails_api_base/pull/444 I missed 2 places that needed the ruby version set


sebastiancaraballo (approved):

.github/workflows/ci.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
+          ruby-version: '3.2.2'
```
santib, line 33: but this wasn't necessary before, right? why is it needed now?

santib: image

I think it's automatically set with the `.ruby-version`

jpascual1994: It's not needed, by default, it uses the `. ruby-version `

guillermoap: I'll remove it from all places and check which version it's using

guillermoap: hadn't refreshed, you are both right

PR #518, opened 2023-11-10
Fix obscure bug

santib:
Having the `status` action defined in `ApiController` which is then inherited by other controllers, was causing a very strange bug related to the [wrap_parameters](https://api.rubyonrails.org/classes/ActionController/ParamsWrapper.html) Rails feature.

In order to reproduce go to the `main` branch, perform the following changes to force the execution of the wrap_parameters feature:
```diff
+++ spec/requests/api/v1/users/update_spec.rb
@@ -9,14 +9,14 @@ describe 'PUT api/v1/user/' do
   before { subject }

   context 'with valid params' do
-    let(:params) { { user: { username: 'new username' } } }
+    let(:params) { { username: 'new username' } }

     it 'returns success' do
       expect(response).to have_http_status(:success)
     end

     it 'updates the user' do
-      expect(user.reload.username).to eq(params[:user][:username])
+      expect(user.reload.username).to eq(params[:username])
     end
```
and run:
```bash
bundle exec rspec spec/requests/api/v1/status_spec.rb spec/requests/api/v1/users/update_spec.rb --order defined
```

as you can see it fails. The reason is that Rails is wrapping the parameters for the projects/update endpoint in a key named "api" instead of the correct one "user"

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -7,19 +7,14 @@ module Api
       include Pundit::Authorization
       include DeviseTokenAuth::Concerns::SetUserByToken
 
+      before_action :authenticate_user!
+
       after_action :verify_authorized, except: :index
       after_action :verify_policy_scoped, only: :index
-
-      before_action :authenticate_user!, except: :status
-      skip_after_action :verify_authorized, only: :status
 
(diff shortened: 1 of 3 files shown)


JulianPasquale (approved):
👀

jpascual1994 (approved):

PerezIgnacio (approved):

PR #527, opened 2023-11-23
Inflect API

santib:
`API` vs `Api`

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Code Climate](https://codeclimate.com/github/rootstrap/rails_api_base/badges/gpa.svg)](https://codeclimate.com/github/rootstrap/rails_api_base)
 [![Test Coverage](https://api.codeclimate.com/v1/badges/63de7f82c79f5fe82f46/test_coverage)](https://codeclimate.com/github/rootstrap/rails_api_base/test_coverage)
 
-Rails Api Base is a boilerplate project for JSON RESTful APIs. It follows the community best practices in terms of standards, security and maintainability, integrating a variety of testing and code quality tools. It's based on Rails 7.0 and Ruby 3.2.
+Rails API Base is a boilerplate project for JSON RESTful APIs. It follows the community best practices in terms of standards, security and maintainability, integrating a variety of testing and code quality tools. It's based on Rails 7.0 and Ruby 3.2.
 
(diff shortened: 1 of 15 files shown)


JulianPasquale (approved):

enzofab91 (approved):

README.md
```diff
@@ -4,7 +4,7 @@
 [![Code Climate](https://codeclimate.com/github/rootstrap/rails_api_base/badges/gpa.svg)](https://codeclimate.com/github/rootstrap/rails_api_base)
 [![Test Coverage](https://api.codeclimate.com/v1/badges/63de7f82c79f5fe82f46/test_coverage)](https://codeclimate.com/github/rootstrap/rails_api_base/test_coverage)
 
-Rails Api Base is a boilerplate project for JSON RESTful APIs. It follows the community best practices in terms of standards, security and maintainability, integrating a variety of testing and code quality tools. It's based on Rails 7.0 and Ruby 3.2.
+Rails API Base is a boilerplate project for JSON RESTful APIs. It follows the community best practices in terms of standards, security and maintainability, integrating a variety of testing and code quality tools. It's based on Rails 7.0 and Ruby 3.2.
```
blacksam07, line 7: With the latest change to include the ability to create fullstack project is possible that we need to update this description

santib: good catch

TimoPeraza:
Just to be curious, why is necessary this change?

megatux:
> > Just to be curious, why is necessary this change?
> 
> Not necessary, I think it's just nicer to use `API` (which is an acronym) instead of `Api`.
> 
> For example Rails uses `API` https://github.com/rails/rails/blob/f4b08457a4ab40418da9aa2155880fea32caf0fe/activemodel/lib/active_model/api.rb#L59.
> 
> But yeah in the end is just a matter of taste. Maybe we can do a poll?
> 
> Vote 👍 for `API` and 👎 for `Api`.

Definitely all in uppercase. There are several examples in the Ruby standard classes. Also, [in Rubocop](https://github.com/rubocop/ruby-style-guide#camelcase-classes). I saw some old posts talking about defining the `ActiveSupport::Inflector` in the Rails init and talking about the convention for the autoloading. We should check if it still applies to Zeitwerk now.

axelpontetto (approved):

PR #536, opened 2023-12-04
Add capybara

sebastiancaraballo:
#### Board:
* [Ticket Notion](https://www.notion.so/rootstrap/Add-feature-system-specs-to-the-admin-site-in-the-Rails-API-Base-just-a-few-489ec65c7c084394918edb1bb4a66045?pvs=4)
---
#### Description:
- Add capybara
- Test admin sign-in

### DIFF
--- a/spec/features/admin/sign_in_spec.rb
+++ b/spec/features/admin/sign_in_spec.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Admin SignIn' do
+  feature 'Signing in' do
+    let(:admin_user) { create(:admin_user) }
+
+    scenario 'Admin signs in successfully' do
+      visit admin_root_path
+      fill_in 'Email', with: admin_user.email
+      fill_in 'Password', with: admin_user.password
+      click_button 'Login'
+
+      expect(page).to have_content 'Signed in successfully.'
+    end
+
+    scenario 'Admin fails to sign in with invalid credentials' do
+      visit admin_root_path
+      fill_in 'Email', with: 'wrong@example.com'
+      fill_in 'Password', with: 'wrongpassword'
+      click_button 'Login'
+
+      expect(page).to have_content 'Invalid Email or password.'
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


sebastiancaraballo:
The tests are failing because there's a problem with a dependency. I resolved this issue locally running `yarn install`. How can I solve this here? Help appreciated cc @santib

spec/features/admin/sign_in_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Admin SignIn' do
+  let(:admin_user) { create(:admin_user) }
+
+  it 'Admin signs in successfully' do
```
enzofab91, line 6: Just a suggestion, what if you put this two its inside a `scenario`?

enzofab91 (approved):
Nice 👏

spec/features/admin/sign_in_spec.rb
sebastiancaraballo: Captura de pantalla 2023-12-04 a la(s) 15 38 50
Rubocop doesn't like using Capybara specific methods

JulianPasquale (approved):

spec/features/admin/sign_in_spec.rb
sebastiancaraballo: Actually, yes. I was [taking a look](https://docs.rubocop.org/rubocop-rspec/cops_rspec_capybara.html#rspeccapybarafeaturemethods) at this and there is a way to enable the usage of Capybara-specific words like `feature`, `scenario`, etc; these are disabled to favor the usage of RSpec-specific words instead. The way I found to allow the usage of Capybara-specific words is adding this config:
```
RSpec/Capybara/FeatureMethods:
  EnabledMethods:
    - feature
    - scenario
    - given
    - background
```
Do we want to enable these words? cc @rootstrap/ruby

santib: Is there any reasoning about RuboCop defaults?

PR #560, opened 2023-12-28
Enhancement/add support js selenium chrome

frankvielma:
#### Board:
 [Notion ticket](https://www.notion.so/rootstrap/add-support-to-run-it-locally-in-non-headless-mode-using-chrome-dc991d84552b49d7bf088108fcc702d7)

---
#### Description:
 * Add support for Javascript in feature specs in non-headless mode using Chrome. 
---
#### Notes:
* Run tests in non-headless mode using:
```console
SHOW=1 rspec spec/
```
---
#### Preview:

[non-headless-chrome.webm](https://github.com/rootstrap/rails_api_base/assets/412081/c5201df0-75e1-4e0c-99f0-6c17b193a0d8)


.github/workflows/ci.yml
```diff
@@ -106,6 +106,12 @@ jobs:
       - name: Get CPU info
         id: cpu_info
         run: echo "cpu_cores=$(nproc)" >> $GITHUB_ENV
+      - name: Install chromedriver
+        uses: nanasess/setup-chromedriver@v2
+      - run: |
+          export DISPLAY=:99
```
santib, line 112: I don't think we need this, we can run it in headless mode, right? 

You might want to check how Rails will do it for new apps here https://github.com/rails/rails/pull/50508/files#diff-dc7109b564892719463be7ade1c27d742684530b8235295c6ed32cb579495712

Gemfile
```diff
@@ -79,6 +79,7 @@ group :test do
   gem 'rspec-openapi', '~> 0.10'
   gem 'rspec-rails', '~> 6.1'
   gem 'rspec-retry', github: 'rootstrap/rspec-retry', branch: 'add-intermittent-callback'
+  gem 'selenium-webdriver', '4.16.0'
```
santib, line 82: use pessimistic versioning `~>`

spec/spec_helper.rb
```diff
@@ -20,6 +20,20 @@
 
 Dir[Rails.root.join('spec/support/**/*.rb')].each { |file| require file }
 
+Capybara.register_driver :chrome do |app|
+  args = %w[no-sandbox disable-gpu]
+  args << (ENV['SHOW'].present? ? 'non-headless' : 'headless')
```
santib, line 25: not sure about the name `SHOW` it's too generic. Maybe something like `HEADLESS` would be better? and we can make it be true by default (in test env) and if we don't want it we do `HEADLESS=false bundle exec rspec`

.github/workflows/ci.yml
frankvielma: :+1:

Gemfile
frankvielma: :+1:

.github/workflows/ci.yml
```diff
@@ -111,7 +111,7 @@ jobs:
           KNAPSACK_CI_NODE_TOTAL: ${{ matrix.ci_node_total }}
           KNAPSACK_CI_NODE_INDEX: ${{ matrix.ci_node_index }}
           PARALLEL_TESTS_CONCURRENCY: ${{ env.cpu_cores }}
-        run: bundle exec parallel_rspec -n $PARALLEL_TESTS_CONCURRENCY -e './bin/parallel_tests'
+        run: NONHEADLESS=false bundle exec parallel_rspec -n $PARALLEL_TESTS_CONCURRENCY -e './bin/parallel_tests'
```
santib, line 114: in the CI we want to run the HEADLESS version, right?

santib (approved):

santib:
Tried running it locally and got this error:
```
F2024-01-04 11:19:57 -0300 Rack app ("GET /admin/login" - (127.0.0.1)): #<ActionView::Template::Error:"Error: Function rgb is missing argument $green.\n        on line 1 of stdin\n>> ay-200{--tw-bg-opacity:1;background-color:rgb(229 231 235/var(--tw-bg-opacit\n   ------------------------------------------^\n">
```

any idea?

spec/spec_helper.rb
```diff
@@ -20,6 +20,20 @@
 
 Dir[Rails.root.join('spec/support/**/*.rb')].each { |file| require file }
 
+Capybara.register_driver :chrome do |app|
+  args = %w[no-sandbox disable-gpu disable-dev-shm-usage]
+  args << (ENV['NONHEADLESS'] == 'false' ? 'headless' : 'non-headless')
```
santib, line 25:
```suggestion
  args << (ENV['HEADLESS'] == 'true' ? 'headless' : 'non-headless')
```

maybe that's simpler? I find the double negation very confusing

frankvielma: :+1:

szcrubyrootstrap (approved):

blacksam07 (approved):

PR #588, opened 2024-01-08
Fullstack script

santib:
This Rails template can be run with `bin/rails app:template LOCATION=./bin/fullstack.rb`

And it will install and configure:
- Turbo
- Stimulus
- Tailwind (via CSSBundling)
- ViewComponent
- Lookbook (previews for ViewComponent)


bin/bootstrap.sh
blacksam07, line 1: We can include here new flags like `--fullstack` to run the `fullstack.rb` task

blacksam07 (approved):
LGTM, amazing 👏🏽

bin/fullstack.rb
```diff
@@ -0,0 +1,115 @@
+# Add gems
+insert_into_file 'Gemfile', after: /gem 'rails'.*\n\n/ do <<-EOF
+gem 'cssbundling-rails', '~> 1.3'
```
JulianPasquale, line 3: I'm wondering if we can avoid specifying the versions here and use the latest at the moment the script is executed.
I get that we want to specify the version on our Gemfile, but I it's gonna be difficult to maintain and upgrade them from this script. Maybe we can tell dependabot to also check this file somehow?

santib: I get your point but I'm not sure what's the best option here.

If we leave the versions unspecified it's possible that the app doesn't work. That's why I decided to follow the same approach as for the app.

That said, I agree this script is going to be hard to keep updated, but not sure what other options we do have. Maybe adding a set of tests that verify that all the installed gems work?

JulianPasquale: Yeah I think that's a good idea, but we can't add those tests directly right? they need to be generated as part of this script 🤔 

Maybe it's ok like this for now, and we can plan to add the tests in the future.

santib: What do you think about this? https://github.com/rootstrap/rails_api_base/pull/588/commits/bef27ba1b42ec6cf6a6856a6b57c7022aece5998

The idea is that if you run the generator it also created an example component and runs the added test to make sure that the whole setup is working

santib: I made some more improvements to make the example component optional (asked in the console) and if accepted along with the example component also create a lookbook preview for it

JulianPasquale: Looks great!! 👏

JulianPasquale (approved):

andresg4 (approved):
👏 🚀

PerezIgnacio (approved):
Awesome 🚀

PR #590, opened 2024-01-10
Add authenticator object

mrodriguez90:
#### Board:
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


app/objects/impersonation/authenticator.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Impersonation
+  class Authenticator
+    def initialize(encrypted_data)
+      @encrypted_data = encrypted_data
+    end
+
+    def build_auth_headers!
+      user.tokens.filter! { |_token, attrs| attrs['impersonated_by'] != admin_user_id }
```
santib, line 10:
```suggestion
      user.tokens.reject! { |_token, attrs| attrs['impersonated_by'] == admin_user_id }
```
really minor comment, feel free to ignore :P

app/controllers/api/v1/impersonates_controller.rb
```diff
@@ -2,13 +2,17 @@
...
-    class UsersController < API::V1::APIController
+    class ImpersonatesController < API::V1::APIController
       skip_before_action :authenticate_user!
       skip_after_action :verify_authorized
 
       def create
-        response.headers.merge!(user.build_auth_headers(token.token, token.client))
+        response.headers.merge!(
+          Impersonation::Authenticator.new(
+            params[:queryParams]
```
santib, line 12:
```suggestion
            params[:auth_enc]
```
?

app/objects/impersonation/authenticator.rb
```diff
@@ -0,0 +1,32 @@
...
+      user.build_auth_headers(token.token, token.client) if user.save!
```
santib, line 11: `user.save!` needs to be called after `user.create_token...` right?

mrodriguez90: I've tested this locally and it seems like it's not necessary.

app/admin/user.rb
```diff
@@ -59,6 +59,6 @@
       user_id: resource.id, admin_user_id: current_admin_user.id
     )
 
-    link_to('Impersonate User', "#{ENV.fetch('FRONTEND_URL')}?queryParams=#{encrypted_data}", method: :get)
+    link_to('Impersonate User', "#{ENV.fetch('FRONTEND_URL')}?auth_enc=#{encrypted_data}", method: :get)
```
lauperalti, line 62: The `auth_enc` name sounds weird to me, what about just `auth_encrypted` 🤔 ?

lauperalti (commented):
🚀 🙌

app/objects/impersonation/authenticator.rb
santib: hmm strange, is there some code saving the user after the create_token method is called then?

app/controllers/api/v1/impersonates_controller.rb
```diff
@@ -2,13 +2,17 @@
 
 module API
   module V1
-    class UsersController < API::V1::APIController
+    class ImpersonatesController < API::V1::APIController
```
lauperalti, line 5: I think we are using `Impersonation` in other classes, does it make sense to name it Impersonation here?

Issue #318, opened 2021-08-27
[FEATURE] Running bootstraper.sh doesn't rename the active_admin site


juan-apa:
**Is your feature request related to a problem? Please describe.**
Running ./bootstraper.sh does not rename the admin panel.

**Describe the solution you'd like**
The name should be changed by the bootstraper in `config/initializers/active_admin.rb`, in the first config called `config.site_title = 'App'`. It could be renamed to `<app_name> Admin`

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #591, opened 2024-01-11
Rename active-admin config file with the project name

blacksam07:
#### Board:
* #318 
* [Notion Ticket](https://www.notion.so/rootstrap/Resolve-https-github-com-rootstrap-rails_api_base-issues-318-92a6ed84051f477789b780ae98aad37b)
---
#### Description:
- Rename ActiveAdmin app name with name received in the `bootstrap.sh` 
---


config/initializers/active_admin.rb
```diff
@@ -6,7 +6,7 @@
   # Set the title that is displayed on the main layout
   # for each of the active admin pages.
   #
-  config.site_title = 'App'
+  config.site_title = 'App Admin'
```
JulianPasquale, line 9: Should this be modified as well? I think you are only searching for "App", right?

blacksam07: @JulianPasquale  According to the ticket, the idea would be to add `Admin` after the `app-name`, which is the reason why I added it here, thoughts?

https://github.com/rootstrap/rails_api_base/issues/318#:~:text=It%20could%20be%20renamed%20to%20%3Capp_name%3E%20Admin

santib: what if we change this to be 
```suggestion
  config.site_title = 'rails_api_base Admin'
```

so that in `bootstrap.sh` we always search and replace `rails_api_base` for the app name. Maybe we can do that in the whole project instead of searching for specific files (e.g. `package.json` would already benefit from that approach)

santib: yeah I like that approach. we just need to check what gets changed but we are probably fine. Let us know how it goes

bootstrap.sh
```diff
@@ -67,8 +67,19 @@ sed_i() {
...
 
-# update project name in database.yml
-sed_i "s/rails_api_base/${project_name}/g" config/database.yml
+# Define the file types to search for
+file_types=("yml" "yaml" "erb" "rb" "json")
+
+for file_type in "${file_types[@]}"
+do
+  # Find and replace the string in files of the current type, and save a backup file
+  find . -name "*.$file_type" -exec sed -i.bak "s|rails_api_base|${project_name}|g" {} +
```
santib, line 76:
```suggestion
  find . -name "*.$file_type" -exec sed -i "" "s|rails_api_base|${project_name}|g" {} +
```

with this change the `.bak` files don't get created and we don't need to remove them below

santib: Actually, I think you need to use the `sed_i` function [defined above](https://github.com/rootstrap/rails_api_base/pull/591/files#diff-99d5a2a8ca4290b61836b3ad37941f2b925e9ff3151c90dcd5ea82364e4122a1L62)

config/initializers/active_admin.rb
blacksam07: I prefer to defined a list of file types to rename and skip issues with some files like (Gemfiles.lock), also I created a backup of each file that then will deleted, this is because it's possible to get an error with reserve files locals in node_modules

bootstrap.sh
santib: cc @guillermoap

santib (approved):
👏

JulianPasquale (approved):

PR #600, opened 2024-01-16
Ensuring consistency of API routes

Zinssmeister:
Small fix on routes file.
Need to force JSON scope to prevent Devise routes from defaulting the request format to `*/*`

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
-  mount_devise_token_auth_for 'User', at: '/api/v1/users', controllers: {
-    registrations: 'api/v1/registrations',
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
-  mount_devise_token_auth_for 'User', at: '/api/v1/users', controllers: {
-    registrations: 'api/v1/registrations',
-    sessions: 'api/v1/sessions',
-    passwords: 'api/v1/passwords'
-  }
+  scope format: 'json' do
```
santib, line 4:
```suggestion
  scope format: :json do
```

santib (approved):
Thanks for debugging the issue and coming up with this fix 👏

Issue #599, opened 2024-01-16
`web` docker image doesn't run


santanaluiz:
## Bug report:
* **Expected Behavior**: all the 3 images should run fine
* **Actual Behavior**: `web` docker compose image doesn't run
* **Steps to Reproduce**: 
  1. Clone project
  2. Run `bootstrap` script
  3. See image failing

* **Version of the repo**: head
* **Ruby and Rails Version**: 7.0.8
* **Rails Stacktrace**: 

Docker logs:

```
2024-01-16 20:22:32 /src/bundle/gems/concurrent-ruby-1.2.2/lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb:14: [BUG] Segmentation fault at 0x0052ffff7f5506c0
2024-01-16 20:22:32 ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [aarch64-linux]
2024-01-16 20:22:32 
2024-01-16 20:22:32 -- Control frame information -----------------------------------------------
2024-01-16 20:22:32 c:0061 p:---- s:0335 e:000334 CFUNC  :resume
2024-01-16 20:22:32 c:0060 p:0007 s:0331 E:001400 BLOCK  /src/bundle/gems/concurrent-ruby-1.2.2/lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb:14 [FINISH]
2024-01-16 20:22:32 c:0059 p:---- s:0328 e:000327 CFUNC  :synchronize
2024-01-16 20:22:32 c:0058 p:0026 s:0324 E:0009d8 METHOD /src/bundle/gems/concurrent-ruby-1.2.2/lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb:12
2024-01-16 20:22:32 c:0057 p:0007 s:0319 e:000318 CLASS  /src/bundle/gems/concurrent-ruby-1.2.2/lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb:18
2024-01-16 20:22:32 c:0056 p:0025 s:0316 e:000315 TOP    /src/bundle/gems/concurrent-ruby-1.2.2/lib/concurrent-ruby/concurrent/atomic/lock_local_var.rb:5 [FINISH]
2024-01-16 20:22:32 c:0055 p:---- s:0313 e:000312 CFUNC  :require
[...]

PR #601, opened 2024-01-17
Updating instructions and docker-compose for running on Docker

santanaluiz:
#### Board:
* #599
---
#### Description:

Updating instructions for running on Docker
---
#### Notes:
* Updated README for clarity
* Added env variable to docker-compose file
* Added env variable to .env.sample file


docker-compose.yml
```diff
@@ -13,6 +13,7 @@ services:
       POSTGRES_USER: ${POSTGRES_USER-postgres}
       POSTGRES_PASSWORD: ${POSTGRES_PASSWORD-postgres}
       POSTGRES_HOST: db
+      SECRET_KEY_BASE: ${SECRET_KEY_BASE-1234567890}
```
JulianPasquale, line 16: Hey @santanaluiz, just curious here, are you running the app in production mode using `docker-compose`?

Our compose file is meant to be used for development purposes only, and I don't think the secret_key_base is needed in those cases.

santanaluiz: I am not using for production, only dev env.
Running rspec fails to me if I don't add it

JulianPasquale: This doesn't sound correct. According to the [Rails docs](https://edgeguides.rubyonrails.org/security.html#session-storage):

> In test and development applications get a secret_key_base derived from the app name

I don't think this is needed to run the tests.

JulianPasquale: This is weird, I cannot reproduce it. I'm actually getting an error but doesn't seem to be related to the secret key

JulianPasquale: @santanaluiz can you please check if this PR fixes your tests https://github.com/rootstrap/rails_api_base/pull/603?

santib:
We just upgraded to Ruby 3.3.1 so the notice is no longer needed. Regarding SECRET_KEY_BASE, seems like it can't be reproduced, so closing the PR for now.

PR #602, opened 2024-01-18
Improvements impersonate

santib:


app/objects/impersonation/verifier.rb
```diff
@@ -0,0 +1,33 @@
...
+        message_verifier.verified(_1, purpose: PURPOSE)
+      }
+    end
+
+    private
+
+    def message_verifier
+      @message_verifier ||= ActiveSupport::MessageVerifier.new(
+        Rails.application.secret_key_base,
+        digest: 'SHA256', serializer: JSON
```
santib, line 29: in Rails 7.1 will be able to add `url_safe: true` option and remove the Base64 stuff.

app/objects/impersonation/verifier.rb
```diff
@@ -0,0 +1,33 @@
...
+module Impersonation
+  class Verifier
+    EXPIRATION = 5.minutes
+    PURPOSE = :impersonation
+
+    def sign!(data)
+      data.then {
+        message_verifier.generate(_1, expires_in: EXPIRATION, purpose: PURPOSE)
+      }.then {
+        Base64.urlsafe_encode64(_1)
```
mrodriguez90, line 12: Is it possible to replace`_1` with `it`?

santib: not yet, we need to wait til Christmas when Ruby 3.4 is going to be released 😅

santib: I ended up removing the `then` chain now that we have only 2 steps

PR #609, opened 2024-01-22
Refactor

santib:


app/objects/impersonation/headers.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Impersonation
+  class Headers
+    def initialize(user, access_token)
+      @user = user
+      @access_token = access_token
+    end
+
+    def build_impersonation_headers
```
mrodriguez90, line 10:
```suggestion
    def build_impersonation_header
```

app/controllers/api/v1/impersonations_controller.rb
```diff
@@ -7,8 +7,12 @@ class ImpersonationsController < API::V1::APIController
       skip_after_action :verify_authorized
 
       def create
+        user, token = Impersonation::Authenticator.new(params[:auth]).authenticate!
+
         response.headers.merge!(
-          Impersonation::Authenticator.new(params[:auth]).build_auth_headers!
+          user.build_auth_headers(token.token, token.client)
+        ).merge!(
+          Impersonation::Headers.new(user, token.token).build_impersonation_headers
```
mrodriguez90, line 15: Why do we need to call this method here? Aren't we also calling this method in `set_headers.rb`?

santib: the thing is that in this controller `current_user` is not set so the concern doesn't add the impersonation header

PR #622, opened 2024-02-06
Mark current_user as impersonated

santib:


app/controllers/api/concerns/impersonation/hooks.rb
```diff
@@ -3,13 +3,15 @@
...
 
         included do
-          after_action do
+          before_action do
+            ::Impersonation::MarkUser.new(current_user, request.headers['Access-Token']).mark!
+
             response.headers.merge!(
-              ::Impersonation::Headers.new(current_user, request.headers['Access-Token']).build_impersonation_header
+              ::Impersonation::Headers.new(current_user).build_impersonation_header
             )
```
santib, line 15: at first I had this in an `after_action` but then I realized it was better to keep it close to the `mark!` since a `user.reload` between the two would clean up the `impersonated_by` attribute

mrodriguez90 (approved):
Nice work, I think we should update the README to share this new capability.

PR #626, opened 2024-02-19
Fix `rspec-rails` gem issues

bigchickenwings:
#### Description:
Currently, cloning this repository, running the setup command, and then running the specs don't work.

#### Arctic Admin related issues (Caused by `rspec-rails`):
The system specs don't work as-is. It throws the following error:
```
ActionView::Template::Error:
   Error: File to import not found or unreadable: arctic_admin/src/scss/main.
             on line 11:1 of app/assets/stylesheets/active_admin.scss
   >> @import 'arctic_admin/src/scss/main';
      ^
# ./app/assets/stylesheets/active_admin.scss:11
# ------------------
# --- Caused by: ---
# SassC::SyntaxError:
#   Error: File to import not found or unreadable: arctic_admin/src/scss/main.
#           on line 11:1 of app/assets/stylesheets/active_admin.scss
#   >> @import 'arctic_admin/src/scss/main';
#      ^
#   ./app/assets/stylesheets/active_admin.scss:11
```

This happens because we are adding the `arctic_admin` through `package.json` instead of adding it as a gem. Therefore, we need to run `bundle exec rails assets:precompile` before running our specs.

If we move the `rspec-rails` gem to both the `:development` and `:test` environments, this will be done automatically for us, as mentioned by @santib 

With these changes, the API will always work on newly-cloned environments, without needing additional setup.

---
#### Before:
Screenshot 2024-02-19 at 12 29 49

#### After:
Screenshot 2024-02-19 at 12 28 59


santib:
Hi @bigchickenwings thanks for the report.

## `spring`
I'm ok removing it. I actually have it completely disabled in my dev env because it brings more problems than solutions to me. If we are going to remove it, there are more places that need to be changed, try doing `git grep -i spring` to find all places. cc @JulianPasquale @guillermoap 

## `arctic_admin`
It works fine for me (and also in the CI). I think you might be missing doing `bundle exec rails assets:precompile` since we are installing `arctic_admin` through the `package.json` instead of the `Gemfile`. It's true that it should be called automatically, but seems that [there is something off](https://github.com/rspec/rspec-rails/issues/2675) cc @PerezIgnacio

bigchickenwings:
Hi, @santib 

Nice catch, I didn't run `bundle exec rails assets:precompile`. Running it does make the specs pass.

Any reason behind this decision?

santib:
> but I do prefer to not have to run bundle exec rails assets:precompile on development and test environments 👀

@bigchickenwings yes, it should be ran automatically. Not sure why it's not happening, maybe it's related to the issue I shared above. But it'd be good to fix it (maybe it's a fix in `rspec-rails`?).

bigchickenwings:
@santib done! Thanks for the help 😄 

Let me know if there's anything else I should address on this PR.

santib:
> @santib done! Thanks for the help 😄
> 
> Let me know if there's anything else I should address on this PR.

Would you mind making 2 separate PRs with the appropriate titles?

One for improving the rspec-rails setup, and the other to remove Spring (also check for more places where it needs to be removed). This way we can review and have conversations on each of them individually and eventually merge them once they are ready without blocking each other.

bigchickenwings:
@santib updated this one to handle `rspec-rails` only.

Will be creating the `spring` one later today 🚀

Gemfile
```diff
@@ -78,8 +79,7 @@ group :test do
   gem 'parallel_tests', '~> 4.5'
   gem 'pg_query', '~> 5.1.0'
   gem 'prosopite', '~> 1.4.2'
-  gem 'rspec-openapi', '~> 0.12'
-  gem 'rspec-rails', '~> 6.1'
+  gem 'rspec-openapi', '~> 0.11'
```
santib, line 82:
```suggestion
  gem 'rspec-openapi', '~> 0.12'
```

santib (approved):

PerezIgnacio (approved):

santib:
Thanks @bigchickenwings 🙌

PR #627, opened 2024-02-19
update bundle version

grosendo2006:
#### Board:
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ ENV GEM_HOME $WORK_ROOT/bundle
 ENV BUNDLE_BIN $GEM_HOME/gems/bin
 ENV PATH $GEM_HOME/bin:$BUNDLE_BIN:$PATH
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,7 +20,7 @@ ENV GEM_HOME $WORK_ROOT/bundle
 ENV BUNDLE_BIN $GEM_HOME/gems/bin
 ENV PATH $GEM_HOME/bin:$BUNDLE_BIN:$PATH
 
-RUN gem install bundler
```
mwolman, line 23: @grosendo2006 This change shouldn't be necessary, cause here:
https://github.com/rootstrap/rails_api_base/blob/f267f15d1aa2b9ca16a7e71937ae580e38134156/Dockerfile#L33
we're already upgrading the bundler version to the latest. WDYT?

PR #634, opened 2024-02-21
Remove `spring` gem to fix gem-related issues

bigchickenwings:
As requested by @santib, I've created this PR for maintainers (and possibly users) to discuss about what to do with `spring` gem.

#### Description:
Currently, cloning this repository, running the setup command, and then running the specs don't work.
### rails db:prepare and related commands issues (Caused by spring):

They are not working in MacOS M1, and are throwing the following error:

```
objc[26168]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
```

This is [a recurring issue, as demonstrated here](https://github.com/rails/rails/issues/38560#issuecomment-983438568).

While we can always just [disable fork safety](https://github.com/rails/rails/issues/38560#issuecomment-590236052), this would go against best practices, and instead we should remove the spring gem or at least disable them for MacOS M1 environments. Let me know the preferred approach.

Also, as [Spring does not offer support for Ruby 3.3 and Rails 7.1](https://github.com/rails/spring?tab=readme-ov-file#compatibility), we should not use it until further updates. As this doesn't really change the functionality of the app, I believe we're good to just remove it for now.

---
#### Notes:
* Let me know if we should disable spring on MacOS M1 environments instead of fully removing it.


JulianPasquale (approved):
👍

rodrieiz (approved):

enzofab91 (approved):

santib:
I was checking and seems that since https://github.com/rails/rails/pull/42997 the new default is not installing Spring. So this PR sounds good to me.

Thanks @bigchickenwings 🙌

PR #674, opened 2024-04-04
Exclude status endpoint to prevent redirect in healthcheck verify

blacksam07:
#### Board:
---
#### Description:
Add `ssl_options` to exclude the status endpoint from ssl redirect and prevent issues with health-check verification 

based on https://api.rubyonrails.org/classes/ActionDispatch/SSL.html

#### Notes:
* Currently we have problem with the Load Balancer when we have the `force_ssl` enabled, with this setup we can omit the ssl redirect in the status endpoint
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -54,6 +54,7 @@ Rails.application.configure do
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
+  config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?('api/v1/status') } } }
 
   # Log to STDOUT by default
   config.logger = ActiveSupport::Logger.new(STDOUT)


config/environments/production.rb
```diff
@@ -54,6 +54,7 @@
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
+  config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?('api/v1/status') } } }
```
JulianPasquale, line 57: This sounds like a very specific use case associated with the infrastructure of your project.
Why we need this now? We were able to deploy many apps with the existing config without any issue.

blacksam07: @JulianPasquale All internal apps that we have disable the `force_ssl` to skip this issue

santib:
I don't get why is this needed. Please explain why someone would hit this endpoint without SSL and why can't it hit it with SSL

blacksam07:
Currently all projects in RS that use a load balancer need to skip the `force_ssl` because breaks the load balancer health check, this issue is because the ELB health-check make the request over the HTTP, if the RAILS app have the force_ssl enable, redirect all HTTP request to the HTTPS and the ELB consider the application as unhealthy because get a 301 status.

another way is keep this setup in any section of the documentation, if we want to keep enable the `force_ssl` enable and have ELB with a health-check we should to include this setup

blacksam07:
@santib after I looking all internal projects that no using `force_ssl` I was assume that we omit this issue and not look solutions about it, I saw this as a solution to have force_ssl enable with a proxy/ELB active and opened this PR 

1. About the infra setup, ELB can sent the HTTPS request but not works properly, continue failing and this setup means maintaining the certificates every x time
2. `assume_ssl` Look great, I didn't see this solution at the moment but we can check in the future

If you are ok, we can close PR

PR #728, opened 2024-05-23
Annotate gem disables Rubocop Layout/LineLength for its comments

DJA89:
#### Board:
-
---
#### Description:

Rubocop will complain if Annotate adds a line that is too long and it will break CI. To fix this I'm making annotate disable Layout/LineLength for its comments.

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/app/models/admin_user.rb
+++ b/app/models/admin_user.rb
@@ -1,28 +1,38 @@
 # frozen_string_literal: true
 
+# rubocop:disable Layout/LineLength
 # == Schema Information
 #
 # Table name: admin_users
 #
-#  id                     :integer          not null, primary key
-#  email                  :string           default(""), not null
+#  id                     :bigint           not null, primary key
+#  provider               :string           default("email"), not null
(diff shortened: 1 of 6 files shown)


app/models/admin_user.rb
```diff
@@ -1,28 +1,38 @@
...
-#  last_sign_in_at        :datetime
-#  current_sign_in_ip     :inet
-#  last_sign_in_ip        :inet
+#  confirmation_token     :string
+#  confirmed_at           :datetime
+#  confirmation_sent_at   :datetime
+#  unconfirmed_email      :string
+#  name                   :string
+#  nickname               :string
+#  image                  :string
```
santib, line 22: I think you are adding columns that don't belong to the rails_api_base?

DJA89: I just ran annotate, didn't think to check the comments, weird that that appeared 🤔. I can delete them by hand, though I'm not sure why they appeared in the first place.

PR #732, opened 2024-05-27
Dockerfile setup improvements

JulianPasquale:
#### Description:
- Update and optimize`Dockerfile` to run in production. It reduces image size by using only necessary dependencies, a ruby-slim image and improves memory management by using jemalloc.
- Add `Dockerfile.dev` for local development. It keeps the Node binaries so we can take advantage of live reloading.
- Improve `docker-compose`. Now we don't need to have use volumes for live reloading as we can use the watch feature, and the web and js processes run in a single service, enabled by foreman.
- Add `docker-compose.test` , configured to run specs with it (has chrome-server service and options needed only in test environment).
- Improve bin scripts to run with docker.
- Add docs to explain new changes and how to use them.
---


.github/workflows/ci.yml
```diff
@@ -26,6 +26,11 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'yarn'
```
santib, line 33: why wasn't this necessary before?

Dockerfile
```diff
@@ -1,45 +1,87 @@
...
-ENV GEM_HOME $WORK_ROOT/bundle
-ENV BUNDLE_BIN $GEM_HOME/gems/bin
-ENV PATH $GEM_HOME/bin:$BUNDLE_BIN:$PATH
+ENV BUNDLE_PATH $APP_HOME/vendor/bundle
+
+# Set prod environment to avoid installing dev dependencies
+ENV BUNDLE_WITHOUT development:test
+ENV BUNDLE_DEPLOYMENT 1
+ENV RAILS_ENV production
+ENV NODE_ENV production
```
santib, line 20: should we specify the yarn version as well?

santib: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt#L38

Dockerfile
```diff
@@ -1,45 +1,87 @@
...
 RUN mkdir -p $APP_HOME
 
-RUN bundle config --path=$GEM_HOME
+# Setup work directory.
+WORKDIR $APP_HOME
+
+# Copy dependencies files and install libraries.
+COPY --link Gemfile Gemfile.lock package.json yarn.lock ./
+
+RUN gem install bundler && bundle install -j 4 && yarn install && \
```
santib, line 42:
```suggestion
RUN gem install bundler && bundle install -j 4 && yarn install --frozen-lockfile && \
```

to make sure the lockfile is respected

bin/docker-entrypoint
```diff
@@ -0,0 +1,14 @@
...
+# If running the rails server then create or migrate existing database
+if ([ "${1}" == "./bin/rails" ] && [ "${2}" == "server" ]) || [ "${1}" == "./bin/dev" ]; then
+  ./bin/rails db:prepare
+fi
+
+if [ "${1}" == "./bin/rspec" ]; then
+  ./bin/rails db:test:prepare
+fi
+
+rm -f tmp/pids/server.pid
```
santib, line 12: why is this needed?

bin/docker-entrypoint
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
```
santib, line 2: do we need to add this jemalloc code?
https://github.com/rails/rails/blob/060d6ca4d111d1e9b14a0a2aa4d3da12718fcb1a/railties/lib/rails/generators/rails/app/templates/docker-entrypoint.tt#L3-L6

bin/helpers.rb
```diff
@@ -19,7 +19,5 @@ def web_service_running?
 end
 
 def web_service
-  abort("\n** ABORTED: docker-compose not installed **") unless docker_compose_installed?
-  out, = Open3.capture2("docker-compose ps --services --filter 'status=running' | grep web")
-  out.strip
+  'web'
```
santib, line 22: ??

docker-compose.yml
```diff
@@ -2,38 +2,37 @@ version: '3.0'
...
-  js:
-    build: .
-    entrypoint: yarn build --watch
-    volumes:
-      - .:/src/myapp
-      - node_modules:/src/myapp/node_modules
+      - 5432:5432
 volumes:
   db_data:
-  node_modules:
```
santib, line 39: just curious but why this removal? wasn't it trying to cache the downloaded node_modules?

docs/docker.md
```diff
@@ -0,0 +1,108 @@
...
+## Intro to bin scripts
+Our scripts in `bin` directory have been updated so we can run them with docker. To do this, you have to set the `DOCKER_ENABLED=true` env variable. This can be done using the .env file, or exporting it in the terminal.
+
+## Intro to Docker
+`Dockerfile.dev` - This Dockerfile is specific for local development purposes. It's optimized to get the live reloading working and assumes the resulting image size is not a concern here.
+
+`Dockerfile` - This Dockerfile is optimized to run in production environments. It ensures a minimal image size to run the app, by installing the minimal needed dependencies, and also adds Jemalloc to improve the Ruby memory management.
+
+## Intro to Docker Compose
+`docker-compose.yml` - This compose file is intented to use only for local development, NOT FOR PRODUCTION. It runs all the services we need in order to run our Rails server, and can be easily extended to include more services if your app needs to. It depends on the `Dockerfile.dev` file and has some "watch" mechanisms to build or reload the app when files change.
```
santib, line 13:
```suggestion
`docker-compose.yml` - This compose file is intended to be used only for local development, NOT FOR PRODUCTION. It runs all the services we need in order to run our Rails server, and can be easily extended to include more services if your app needs to. It depends on the `Dockerfile.dev` file and has some "watch" mechanisms to build or reload the app when files change.
```

docs/docker.md
```diff
@@ -0,0 +1,108 @@
...
+> [!WARNING]
+> By default this command will not stop the db and chrome-server services when the process completes. To achieve this you can run `./bin/rspec --rm` instead
+
+### How to use Rails console
+```bash
+./bin/rails c
+```
+
+### How to start the Rails server
+Currently there is no script for doing this with docker. Check how to start the Rails server using Docker Compose in the next section.
```
santib, line 35: but running `./bin/rails s` will use Docker if I have the ENV VAR and Docker running because of https://github.com/rootstrap/rails_api_base/pull/732/files#diff-f7968f0a612c299c9774edbd1a3c2169dbc5a1809ba9208ccc50beef597364b3 right?

.github/workflows/ci.yml
JulianPasquale: Before these changes, we did not have a Node version specified. This PR fixes the node version in the `package.json` and also adds a `.nvmrc` file.

The [ubuntu image](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) that Github Actions uses already includes a Node version by default, so we were using that one, but right not it doesn't match with the one we set and that's why we need to install it manually

Dockerfile
JulianPasquale: hmm good point. I think we should either specify the version in the Dockerfile or not specify a version number at all.

bin/docker-entrypoint
JulianPasquale: We are already doing it in the Dockerfile: https://github.com/rootstrap/rails_api_base/pull/732/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R75

I think it makes more sense to have it as a `RUN` step (build time) than moving it to the `ENTRYPOINT` command (Run time). Just checked and the env variable is correctly setup on runtime

bin/helpers.rb
JulianPasquale: The main version of this script is getting the list of running services using `docker-compose` and re-using the one called "web" (or including "web" in its name). This basically means that in order to run any of our `bin` scripts, you had to manually start all the services with `docker-compose`.

We are removing that dependency. Now we don't check if there is an existing "web" process, we will always be creating new containers from the same image.

docker-compose.yml
JulianPasquale: Two things to note here:

1 - We are using the [docker-compose watch](https://docs.docker.com/compose/file-watch/) feature that re-builds the image if the `package.json` file is updated
```YAML
develop:
  watch:
    - action: rebuild
      path: package.json
```

2 - The `dockerfile.dev` file has `CMD ["./bin/dev"]`. This means that by default (and we are not overriding this in the compose file) it will run that script when you start a container from the built image. This scripts uses foreman and run both the Rails server and a node watch script.


So the flow would be like:
1. First time you build and run the compose services: Image is created using the `dockerfile.dev` file and both Rails and Node watch process are executed
2. If you update your dependencies on the `package.json` file, step 1 is executed again, so new/updated libraries are refreshed and the watch process updates the JS code.

docs/docker.md
JulianPasquale: Yes, that's correct. The issue is that the resulting command will be
```bash
docker compose run web ./bin/rails s
```

There is a difference between doing a `run` and doing `up`.
When you use `up` all the services are executed and all the ports are exported. If you use `run <service>` instead, the service is started (in this case the web service) along with all its dependencies (servicies specified in the `depends_on` attribute) but the ports are never linked. So you are going to see you server started, but hitting the port `3000` on your browser will just no work.

JulianPasquale: Actually I was just reading the [run command docs](https://docs.docker.com/reference/cli/docker/compose/run/#description). Looks like we can use the `--service-ports` flag to actually export them... Maybe it would be nice to test it and see if there are no collisions. cc: @PerezIgnacio @smerlo @axelpontetto

bin/docker-entrypoint
JulianPasquale: It's probably not 🙃

.github/workflows/ci.yml
santib: ahhh nice

bin/docker-entrypoint
santib: ahhh ok great. Maybe you have a PR for Rails then 😉

bin/helpers.rb
santib: ok, I guess this method can go away and just use the string `web` in `bin/web` and remove `&& !web_service.empty?` from line 18 in this file

santib (approved):
🌟 this is awesome

Dockerfile.dev
```diff
@@ -0,0 +1,46 @@
...
+
+# Use Node image so we can pull the binaries from here.
+FROM node:$NODE_VERSION as node
+
+# Ruby build image.
+FROM ruby:${RUBY_VERSION}-slim
+
+RUN apt-get update -qq && \
+    apt-get install -y build-essential libssl-dev libpq-dev vim git libsasl2-dev && \
+    rm -rf /var/lib/apt/lists/*
```
lauperalti, line 12: Do we have Specific versions for the apt-get packages or not needed?

Dockerfile.dev
```diff
@@ -0,0 +1,46 @@
...
+ENV LANG C.UTF-8
+ENV BUNDLE_PATH $WORK_ROOT/bundle
+
+# Create app directory.
+RUN mkdir -p $APP_HOME
+
+# Setup work directory.
+WORKDIR $APP_HOME
+
+RUN gem install foreman
```
lauperalti, line 30: Is it possible to combine this statement with `RUN gem install bundler ...`?

lauperalti (approved):
Love this 💪 ! Excellent documentation and step by step guide.

Dockerfile.dev
smerlo: It's not needed to specify the versions for these packages. Usually these packages are provided by the SO

PR #733, opened 2024-05-31
Config ActiveStorage Current without using concern

andresg4:
#### Board:
---
#### Description:
This PR is for having the fix for the issue when using Active Storage Disk Service in development and test environments but without using the `ActiveStorage::SetCurrent` concern. Also this fix applies now for `APIController`. 

This solution was seen here https://github.com/rails/rails/issues/40855#issuecomment-1970876302

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/lib/active_storage/service/disk_with_host_service.rb
+++ b/lib/active_storage/service/disk_with_host_service.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'active_storage/service/disk_service'
+
+module ActiveStorage
+  class Service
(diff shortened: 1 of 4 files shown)


lib/active_storage/service/disk_with_host_service.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'active_storage/service/disk_service'
```
santib, line 3: Is this require needed?

santib (approved):

JulianPasquale (approved):

PR #754, opened 2024-06-12
Fix Docker usage

### LINKED ISSUE #753: Rails console not working without Docker
Can't open the rails console when not using Docker

image

santib:
Now, if `Docker` is off, rails commands will still work. Fixes #753


JulianPasquale (approved):
🏅

PR #765, opened 2024-06-24
Upgrade `ruby` and `bundler` versions

mwolman:
#### Description:
Upgrade `ruby` and `bundler` versions to the latest


Gemfile.lock
```diff
@@ -598,7 +598,8 @@ GEM
     zeitwerk (2.6.16)
 
 PLATFORMS
-  ruby
+  arm64-darwin-23
```
JulianPasquale, line 601: We should probably keep `ruby` as well

Gemfile.lock
```diff
@@ -330,7 +330,6 @@ GEM
     matrix (0.4.2)
     method_source (1.1.0)
     mini_mime (1.1.5)
-    mini_portile2 (2.8.7)
```
JulianPasquale, line 333: This is weird... Do you know why is being removed?

mwolman: I have no idea 🤷🏻‍♀️ both the `ruby` and this line were removed when I ran `bundle _2.5.14_ install`

Gemfile.lock
mwolman: This line was removed when I ran `bundle _2.5.14_ install` 🤔

JulianPasquale: Yeah, It's probably due to your system architecture. We have to make sure it works for all of them, so I say let's add it back manually

mwolman: It's the same problem with `nokogiri`. I see now that it's installing `nokogiri (1.16.6-arm64-darwin)`. But I don't know how to fix it cause if I add the `ruby` line manually when I run `bundle install,` it removes the line again 🥲

JulianPasquale: Yeah but we still need to fix this. I tried building the Dockerfile locally and got a error due to a missing platform
image

Running `bundle lock --add-platform ruby` adds back the ruby platform and the Docker build starts to work again. I also manually removed the other platforms and only left `ruby` and it works on my local env as well. I think that's probably the best workaround

Gemfile.lock
JulianPasquale: After adding back the platform, this gem was added back to my Gemfile.lock file. I think it's still a dependency but it's not needed on your local platform. We have to make sure we keep it.

JulianPasquale (approved):

santib (approved):

PR #772, opened 2024-07-05
Change heroku deployment to use the dockerfile

axelpontetto:
#### Board:
* [Ticket #NUMBER_OF_THE_TICKET](https://www.notion.so/rootstrap/One-off-Tasks-e7bed551a599475ea88f9c825c9cfdb7?p=24a4253f537c4617b73fbb8e5148ba1e&pm=s)
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/heroku.yml
+++ b/heroku.yml
@@ -1,0 +1,13 @@
+setup:
+  addons:
+    - plan: heroku-postgresql
+      as: DATABASE
+build:
+  docker:
+    web: Dockerfile
+run:
+  web: bundle exec puma -C config/puma.rb
+  worker:
+    command:
+      - ./bin/delayed_job start
+    image: web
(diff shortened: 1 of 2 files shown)


JulianPasquale (approved):

heroku.yml
```diff
@@ -0,0 +1,13 @@
+setup:
+  addons:
+    - plan: heroku-postgresql
+      as: DATABASE
```
santib, line 4: I think this is the default

santib (approved):

frankvielma (approved):

PerezIgnacio (approved):

heroku.yml
axelpontetto: yes, do you think it's best to remove it?

santib: Usually, I don't set default values, but I'm ok either way

PR #774, opened 2024-07-08
Fix fullstack script and Rails generators with Docker

JulianPasquale:
- Remove the `BUNDLE_PATH` env variable from `Dockerfile.dev`. This variable was changing the path where bundler is installing the gems and slowing the volume mount Docker process. Also generating some conflicts with the `fullstack.rb` script and gems were installed twice, in different locations.
- Update the entrypoint script so it runs the `db:test:prepare` command when `RAILS_ENV=true`
- Replace the Docker Compose watch feature with a bind mount volume so Rails generators and scripts are able to create files on the host
    - The watch mechanism does nothing if the path it is watching is already included in a volume. `WARN[0064] path 'Gemfile' also declared by a bind mount volume, this path won't be monitored! `

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -20,7 +20,6 @@ COPY --from=node /opt /opt
 ENV WORK_ROOT /src
 ENV APP_HOME $WORK_ROOT/app/
 ENV LANG C.UTF-8
-ENV BUNDLE_PATH $WORK_ROOT/bundle
 
 # Create app directory.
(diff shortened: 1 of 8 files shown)


axelpontetto:
looks good to me!

axelpontetto (approved):

santib (approved):

PerezIgnacio (approved):
Looks great! Can we also update the docs were we mention watch feature?

PR #807, opened 2024-08-14
Remove sass-rails gem and use cssbundling and dartsass instead

andresg4:
#### Board:
---
#### Description:

The `sass-rails` gem is deprecated, in this PR we use [cssbundling-rails](https://github.com/rails/cssbundling-rails) and configure `dartsass` with it

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


Gemfile.lock
```diff
@@ -605,8 +594,8 @@ GEM
 
 PLATFORMS
   arm64-darwin-22
+  arm64-darwin-23
   ruby
-  x86_64-linux
```
santib, line 609: I don't think we want to remove x86

package.json
```diff
@@ -1,20 +1,22 @@
...
   },
   "dependencies": {
     "@activeadmin/activeadmin": "3.2.3",
     "@fortawesome/fontawesome-free": "6.6.0",
     "arctic_admin": "4.3.1",
     "esbuild": "0.23.0",
     "jquery": "3.7.1",
-    "jquery-ui": "1.14.0",
-    "jquery-ujs": "1.2.3"
+    "jquery-ui": "1.13.3",
```
santib, line 14: keep the latest version

santib (approved):
Niceeeee

Gemfile
```diff
@@ -30,9 +31,8 @@ gem 'pundit', '~> 2.3'
 gem 'rack-cors', '~> 2.0'
 gem 'rswag-api', '~> 2.13.0'
 gem 'rswag-ui', '~> 2.13.0'
-gem 'sass-rails', '~> 6.0.0'
 gem 'sendgrid', '~> 1.2.4'
-gem 'sprockets', '~> 4.2.1'
+gem 'sprockets-rails', '~> 3.5', '>= 3.5.2'
```
JulianPasquale, line 35: Why is this changing from `sprockets` to `sprokets-rails`?

package.json
```diff
@@ -1,20 +1,22 @@
...
+    "jquery-ujs": "1.2.3",
+    "sass": "^1.77.8"
   },
   "scripts": {
-    "build": "node esbuild.config.mjs"
+    "build": "node esbuild.config.mjs",
+    "build:css": "sass ./app/assets/stylesheets:./app/assets/builds --no-source-map --load-path=node_modules"
```
JulianPasquale, line 20: Shouldn't the `build` script run this as well? Maybe we can call this `prebuild` so it runs automatically?

PerezIgnacio (approved):
Looks good! Do we need any updates on fullstack script? (which installs tailwind and replaces build:css)

Gemfile
andresg4: if we keep the `sprockets` gem and remove the `sass-rails`, we get this error `'require': cannot load such file -- sprockets/railtie`. Replacing it with the `sprockets-rails` one solves it. I'm not really sure why this happens

JulianPasquale: Would be nice to understand why and what is the impact of this change. I assume we're good, but let's validate it doesn't have any negative impact

santib: just in case, `sprockets-rails` was already being installed as a dependency on `sass-rails`. https://github.com/rootstrap/rails_api_base/pull/807/files#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fL571-L573

Basically our previous setup wasn't so good, that's the whole issue here. `sprockets-rails` is required for our app that we are using the traditional asset pipeline.

package.json
santib: I think this was added by the cssbundling gem. I mean, the default behavior.

JulianPasquale: Yeah, I'm just saying that would be nice to have something like
```JSON
scripts: {
  "build": "yarn build:js && yarn build:css",
  "build:js": "node esbuild.config.mjs",
  "build:css": "sass ./app/assets/stylesheets:./app/assets/builds --no-source-map --load-path=node_modules"
}
```

That way `build` runs both process. Or another option would be to rename the `build:css` script to `prebuild` so it's always executed when we call `build`. (I like the first option because it's more flexible tbh).

Anyway, if you want to keep it this way, we might need to add the `build:css` step to our `Dockerfile.dev`

bin/fullstack.rb
```diff
@@ -22,6 +21,11 @@
 # Run CSS generator
 rails_command 'css:install:tailwind'
 
+# Add Sass build for Active Admin styles
+insert_into_file 'package.json', after: "application.css --minify" do
+  ' && sass ./app/assets/stylesheets/active_admin.scss:./app/assets/builds/active_admin.css --no-source-map --load-path=node_modules'
```
santib, line 26: it'd be better to have
`sass ./app/assets/stylesheets:./app/assets/builds --no-source-map --load-path=node_modules`

santib (approved):

JulianPasquale (approved):

jpascual1994 (approved):

PR #808, opened 2024-08-16
Adjust fullstack script with Docker

andresg4:
#### Board:
---
#### Description:

- Added the `volume` to the `docker-compose.test.yml` so the changes from the `fullstack` script are applied to the local files and not only in the container
- Added `&& docker compose -f docker-compose.test.yml down` to the fullstack command in the docs so the `tests` container is stopped when it finishes running

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -20,6 +20,8 @@ services:
       RAILS_ENV: test
     tty: true
     stdin_open: true
+    volumes:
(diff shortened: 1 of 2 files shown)


JulianPasquale (approved):
👏

santib (approved):

PR #828, opened 2024-09-03
Upgrade node and yarn versions

PerezIgnacio:
#### Board:
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "rails_api_base",
-  "private": "true",
+  "private": true,
   "engines": {
-    "node": "20.10.0",
-    "yarn": "1.22"
+    "node": "22.7.0",
```
santib, line 5: should we make this a little more flexible? something like what in Ruby would be `~22.0` ? So that as long as the major version matches we are good?

yarn.lock
```diff
@@ -1,310 +1,1330 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
santib, line 1: we went from ~300LOC to ~1600LOC 🤯 

I see new dependencies, that I guess aren't needed? 

is it because of `"packageManager": "yarn@4.4.1",` ?

PerezIgnacio: yes, it seems to be adding some dev dependencies, not sure exactly why yet

PerezIgnacio: This happened due to a recurrent issue with fsevents and yarn, which ends up installing node-gyp even if it's not listed as dependency. sass package depends on chokidar, which depended on fsevents. Coincidentally, few days ago they removed this dependency from chokidar, so upgrading sass solves this. 😄

package.json
```diff
@@ -13,7 +14,7 @@
     "jquery": "3.7.1",
     "jquery-ui": "1.14.0",
     "jquery-ujs": "1.2.3",
-    "sass": "^1.78.0"
+    "sass": "^1.79.3"
   },
   "scripts": {
```
JulianPasquale, line 19: Can we add a pre-install script to automatically run `corepack enable`? That way we don't have to do it manually anywhere
```suggestion
  "scripts": {
    "preinstall": "corepack enable",
```

JulianPasquale: https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts

PerezIgnacio: I like the idea, but won't you still need to have corepack enabled/yarn installed to run this?

santib (approved):

axelpontetto (approved):

package.json
PerezIgnacio: Not really. If corepack is enabled, it will catch any yarn command and try to install it (based on the version from packageManager in package.json). Before, we needed to have yarn already installed to run any script, now we need to have corepack enabled (which is already included in nodejs releases).

JulianPasquale (approved):

PR #837, opened 2024-09-16
Update dockerfile to use a rootless user

axelpontetto:
#### Board:
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


Dockerfile
```diff
@@ -61,19 +61,32 @@ RUN apt-get update -qq && \
...
-COPY --from=builder $APP_HOME/tmp/ $APP_HOME/tmp/
-COPY --from=builder $APP_HOME/vendor/ $APP_HOME/vendor/
+COPY --link --chown=$USERNAME:$USERNAME --chmod=700 . .
+COPY --from=builder --chown=$USERNAME:$USERNAME --chmod=700 $APP_HOME/public/ $APP_HOME/public/
+COPY --from=builder --chown=$USERNAME:$USERNAME --chmod=700 $APP_HOME/tmp/ $APP_HOME/tmp/
+COPY --from=builder --chown=$USERNAME:$USERNAME --chmod=700 $APP_HOME/vendor/ $APP_HOME/vendor/
 
+USER root
 RUN ln -s /usr/lib/*-linux-gnu/libjemalloc.so.2 /usr/lib/libjemalloc.so.2
+USER $USERNAME
```
JulianPasquale, line 89: Have you tried moving this line up? I assume if we login as that username before running all the copy, we wouldn't need to specify the owner in the command

Dockerfile
```diff
@@ -61,19 +61,32 @@ RUN apt-get update -qq && \
     apt-get install --no-install-recommends -y curl libpq-dev libvips libjemalloc2 && \
     apt-get clean
 
+ENV USERNAME rails_api_base
```
JulianPasquale, line 64: Maybe this way we won't need to change if for all the projects 🙂 
```suggestion
ENV USERNAME rails
```

Dockerfile
```diff
@@ -61,19 +61,32 @@ RUN apt-get update -qq && \
...
+ENV USER_UID 1000
+ENV USER_GID 1000
```
JulianPasquale, line 66: Should we use `ENV` or `ARG` here? 🤔 https://docs.docker.com/build/building/variables/

axelpontetto: At first I tried using ARG but I remember I had some issues with that. I will take another look.

Dockerfile
axelpontetto: Yes, I already tried that. Not sure if there is another workaround.

PR #854, opened 2024-10-09
Add health check on docker compose

antoalconti:
#### Board:
* [Ticket #NUMBER_OF_THE_TICKET](https://www.notion.so/rootstrap/One-off-Tasks-e7bed551a599475ea88f9c825c9cfdb7?p=866941cff01349f3af35d61f93e315e8&pm=s)
---
#### Description:

--- Checks for web and db were added, now on status we can chech if healthy or not
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---
#### Preview:
image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,12 @@ services:
       - db
     links:
       - db:db
+    healthcheck:
+        test: ["CMD", "curl", "-f", "http://localhost:3000"]
+        interval: 1m30s
+        timeout: 10s
+        retries: 3
+        start_period: 40s
   db:
     image: postgres:15
     environment:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -18,6 +18,12 @@ services:
       - db
     links:
       - db:db
+    healthcheck:
+        test: ["CMD", "curl", "-f", "http://localhost:3000"]
+        interval: 1m30s
+        timeout: 10s
+        retries: 3
+        start_period: 40s
```
JulianPasquale, line 26: I feel these values are a bit high 🤔 does it really take 40s for the server to start?

antoalconti: i use default values, not sure which are the average, maybe 20 seconds? @JulianPasquale

axelpontetto (approved):

santib (approved):

docker-compose.yml
santib: where did you take the defaults from?

antoalconti: https://docs.docker.com/reference/compose-file/services/#healthcheck

PR #865, opened 2024-10-16
Impersonation prevent the redirect when admin session expired

blacksam07:
#### Description:
Prevent the error when the admin user still in the show page and doing click in the impersonate button

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
* LOW
---

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -3,6 +3,13 @@
 ActiveAdmin.register User do
   permit_params :email, :first_name, :last_name, :username, :password, :password_confirmation
 
+  member_action :impersonate, method: :post do
+    signed_data = Impersonation::Verifier.new.sign!(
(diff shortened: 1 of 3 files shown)


mrodriguez90 (approved):
Nice work! thanks

app/admin/user.rb
```diff
@@ -3,6 +3,13 @@
 ActiveAdmin.register User do
   permit_params :email, :first_name, :last_name, :username, :password, :password_confirmation
 
+  member_action :impersonate, method: :post do
```
JulianPasquale, line 6: Should we also check for `ENV['IMPERSONATION_URL']` before defining this action?

JulianPasquale (approved):

santib (approved):

PR #888, opened 2024-11-08
Improve CD workflow

blacksam07:
#### Board:
* [Ticket CD process improvements](https://www.notion.so/rootstrap/CD-process-improvements-12d59347eba680c3ba21c2302008b19c?pvs=4)
---
#### Description:
* Adds new steps in the CD workflow to get all `backend` defined parameters in AWS Parameter Store, format them, and pass them to the definition task.
* Retrieve parameters in base of the name convention

---
#### Notes:

---
#### Tasks:
---
#### Risk:
* Not tested in real environment
---
image

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -36,7 +36,9 @@ jobs:
         with:
           context: .
           push: true
-          tags: ${{ steps.login-ecr.outputs.registry }}/${{ vars.ECR_REPOSITORY }}:${{ github.sha }}
+          tags: |
+            ${{ steps.login-ecr.outputs.registry }}/${{ vars.ECR_REPOSITORY }}:${{ github.sha }}
+            ${{ steps.login-ecr.outputs.registry }}/${{ vars.ECR_REPOSITORY }}:latest
 
       - name: Get the image digest
         id: image-digest
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Download Parameter Store Values
         id: ssm-download
         run: |
-          PARAMETERS_JSON=$(aws ssm describe-parameters --query "Parameters[*].{Name:Name,ARN:ARN}" --output json | jq -c '.')
+          PARAMETERS_JSON=$(aws ssm describe-parameters --query "Parameters[?contains(Name, 'backend')].{Name:Name,ARN:ARN}" --output json | jq -c '.')
           echo "parameters=${PARAMETERS_JSON}" >> $GITHUB_OUTPUT
 
       - name: Format SSM Parameters
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -3,9 +3,9 @@ name: CD
 on:
   workflow_dispatch:
   #   push:
-  # Uncomment and add the necessary branches to enable automatic deployment on AWS
-  # branches:
-  # - main
+      # Uncomment and add the necessary branches to enable automatic deployment on AWS
+      # branches:
+        # - main
 
 jobs:
   deploy:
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_01_09_165800) do
+ActiveRecord::Schema[7.2].define(version: 2024_01_09_165800) do
```
JulianPasquale, line 13: We should rollback this change, just to make sure nothing else is updated unintentionally here

.github/workflows/cd.yml
```diff
@@ -46,13 +46,29 @@ jobs:
         run: |
           aws ecs describe-task-definition --task-definition ${{ vars.ECS_TASK_DEFINITION }} --query taskDefinition > ${{ vars.ECS_TASK_DEFINITION_PATH }}
 
+      - name: Download Parameter Store Values
+        id: ssm-download
+        run: |
+          PARAMETERS_JSON=$(aws ssm get-parameters --query "Parameters[*].{Name:Name,ARN:ARN}" --output json | jq -c '.')
```
JulianPasquale, line 52: I think we need to create a name convention here so we only pull the secrets that belong to the Rails app (assuming the same AWS account can store secrets for other apps like the FE app)

blacksam07: Sure, this is one of the posible changes that we need to include, another is update the `format_secrets.rb` in case that we need to transform the variable name that we receive in the response

bin/format_secrets.rb
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
+require 'json'
+
+data = JSON.parse(ENV['SSM_PARAMETERS_JSON'])
```
JulianPasquale, line 4: What do you think is better here, using an env var or just receiving the json as a parameter in the script?

.github/workflows/cd.yml
```diff
@@ -46,13 +48,29 @@ jobs:
         run: |
           aws ecs describe-task-definition --task-definition ${{ vars.ECS_TASK_DEFINITION }} --query taskDefinition > ${{ vars.ECS_TASK_DEFINITION_PATH }}
 
+      - name: Download Parameter Store Values
+        id: ssm-download
+        run: |
+          PARAMETERS_JSON=$(aws ssm describe-parameters --query "Parameters[?contains(Name, 'backend')].{Name:Name,ARN:ARN}" --output json | jq -c '.')
```
JulianPasquale, line 54: Shoulds we use `contains` or maybe only check if it starts with that key? Just thinking that maybe some other keys are named backend and we end up pulling them as well

docs/cd_with_aws.md
```diff
@@ -34,7 +37,28 @@ Before you start, make sure you have the following:
...
+   - **Access AWS Systems Manager**:
+
+     - Log in to your AWS console.
+     - Navigate to **Systems Manager** and select **Parameter Store**.
+
+   - **Create Parameters**:
+
+     - Click **Create parameter**.
+     - Fill out the details:
+       - **Name**: Provide a unique and valid name (e.g., `/rails_api_base/backend/service_api_key`).
```
JulianPasquale, line 51: I think we should also clarify what's the expected naming convention here

bin/format_secrets.rb
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
```
santib, line 1: I think the name of the file should be more explicit that it's formatting secrets only the AWS on the CD flow. It's not a generic formatting script

bin/format_secrets.rb
blacksam07: Could be work as well, I can include the change to receive the json as parameter

.github/workflows/cd.yml
blacksam07: I use `contains` because the name convention that Devops use is `project_name/(backend|frontend)/var_name` and I think is a good way to search, which scenery do you think could be retrieve keys that I don't want

docs/cd_with_aws.md
blacksam07: Sure

.github/workflows/cd.yml
```diff
@@ -3,9 +3,9 @@ name: CD
 on:
   workflow_dispatch:
   #   push:
-      # Uncomment and add the necessary branches to enable automatic deployment on AWS
-      # branches:
-        # - main
+  # Uncomment and add the necessary branches to enable automatic deployment on AWS
+  # branches:
+  # - main
```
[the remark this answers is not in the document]
blacksam07: Sure I will remove, I have a troubles with the editor because automatically changes these lines

JulianPasquale (approved):
👏

santib (approved):

PR #894, opened 2024-11-12
Colorize logs in dev

santib:
Taken from https://gist.github.com/kyrylo/3d90f7a656d1a0accf244b8f1d25999b

Improves logs legibility by colorizing them by the log level

Another additional gem that may be worth adding (but requires more manual setup) is https://github.com/mechanicles/colorize_logs

BEFORE:
image

AFTER:
image

### DIFF
--- a/config/initializers/colorized_logger.rb
+++ b/config/initializers/colorized_logger.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+# This initializer adds color to the Rails logger output. It's a nice way to
+# visually distinguish log levels.
+module ColorizedLogger
+  COLOR_CODES = {
+    debug:   "\e[36m",  # Cyan
+    info:    "\e[32m",  # Green
+    warn:    "\e[33m",  # Yellow
+    error:   "\e[31m",  # Red
+    fatal:   "\e[35m",  # Magenta
+    unknown: "\e[37m"   # White (or terminal default)
+  }.freeze
+
+  RESET = "\e[0m"
+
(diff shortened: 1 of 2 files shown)


JulianPasquale (approved):
😎

enzofab91 (approved):
Nice!

config/initializers/colorized_logger.rb
```diff
@@ -0,0 +1,50 @@
...
+    COLOR_CODES = {
+      debug: "\e[36m", # Cyan
+      info: "\e[32m",  # Green
+      warn: "\e[33m",  # Yellow
+      error: "\e[31m",  # Red
+      fatal: "\e[35m",  # Magenta
+      unknown: "\e[37m" # White (or terminal default)
+    }.freeze
+
+    RESET = "\e[0m"
```
blacksam07, line 16:
```suggestion
    RESET = "\e[0m".freeze
```

blacksam07 (commented):
LGTM, only include a freeze for a constant 

and just only for fun you can use `define_method` to create all repetitive methods
```
    COLOR_CODES.keys.each do |level|
      define_method(level) do |progname = nil, &block|
        super(colorize(level, progname, &block))
      end
    end
```

config/initializers/colorized_logger.rb
santib: strings are frozen by default because of the magic comment at the top of the file
```
# frozen_string_literal: true
```

mrodriguez90 (approved):
🌈 🌈

PR #933, opened 2024-12-19
Remove Oj

santib:
Once we upgrade to Ruby 3.4 Ruby native JSON implementation will receive many performance improvements. That said, I don't think `oj` should be here by default, so I'd remove it right now and not wait until christmas.

https://github.com/ruby/json/blob/master/CHANGES.md

https://byroot.github.io/ruby/json/2024/12/15/optimizing-ruby-json-part-1.html


enzofab91 (approved):
🧹

andresg4 (approved):

PR #937, opened 2024-12-26
Upgrade to Ruby 3.4.1

enzofab91:
#### Board:
---
#### Description:

Upgrade to [Ruby 3.4.1](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-1-released/) released today.

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---


Gemfile.lock
```diff
@@ -265,12 +263,6 @@ GEM
     google-protobuf (4.29.2)
       bigdecimal
       rake (>= 13)
-    google-protobuf (4.29.2-arm64-darwin)
-      bigdecimal
-      rake (>= 13)
-    google-protobuf (4.29.2-x86_64-linux)
-      bigdecimal
-      rake (>= 13)
```
santib, line 273: hmmm strange that these gems for these platforms got removed 🤔

Gemfile
```diff
@@ -2,7 +2,7 @@
 
 source 'https://rubygems.org'
 
-ruby '~> 3.3.5'
+ruby '~> 3.4.1'
```
santib, line 5: WDYT about
`ruby file: '.ruby-version'` ?

https://bundler.io/guides/gemfile_ruby.html
https://andycroll.com/ruby/read-ruby-version-in-your-gemfile/

santib (approved):
👏

Gemfile.lock
enzofab91: This is odd. I try to revert this changes but `bundle install` keeps doing it.

When it fails, on [GH actions](https://github.com/rootstrap/rails_api_base/actions/runs/12509037408/job/34897961331) it says :
```
ffi-1.17.0-x86_64-linux-gnu requires ruby version < 3.4.dev, >= 2.5, which is
incompatible with the current version, 3.4.1
```

Maybe is no more needed? I couldn't find anything on google 😂

JulianPasquale (approved):

Gemfile.lock
santib: 😮 ok. Let's go ahead and if something breaks we can fix it

PR #938, opened 2024-12-30
Use Puma command to start the Rails server

JulianPasquale:


santib (approved):

bin/docker-entrypoint
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 # If running the rails server then create or migrate existing database
-if ([ "${1}" == "./bin/rails" ] && [ "${2}" == "server" ]) || [ "${1}" == "./bin/dev" ]; then
+if ([ "${1}" == "./bin/rails" ] && [ "${2}" == "server" ]) || [ "${1}" == "./bin/dev" ] || [[ $@ == *"bundle exec puma"* ]]; then; then
```
hoangbap010595, line 4: Fixed syntax error in docker-entrypoint script:
- Removed extra semicolon that was causing "syntax error near unexpected token ';'"
- The error occurred in the if condition that checks for Rails server and Puma execution

```if ([ "${1}" == "./bin/rails" ] && [ "${2}" == "server" ]) || [ "${1}" == "./bin/dev" ] || [[ $@ == *"bundle exec puma"* ]]; then```

santib: Good finding! Would you like to open a PR fixing it?

PR #993, opened 2025-03-05
Update users_controller.rb

gautham-nambi-1010413:
#### Board:
---
#### Description:

---
#### Notes:
*
---
#### Tasks:
---
#### Risk:
*
---

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -16,7 +16,7 @@ module API
       private
 
       def update_user_params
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -16,7 +16,7 @@ def update
       private
 
       def update_user_params
-        params.expect(user: %i[username first_name last_name email])
+        params.expect(user: %i[first_name last_name email])
```
santib, line 19: why do you think we should prevent users from changing their `username`?

PR #1002, opened 2025-03-11
Use release script when deploying with containers

JulianPasquale:
#### Description:
When using the `heroku.yml` to deploy apps with the container stack on Heroku, the release script is not being executed. This PR fixes the issue by adding the script into the `heroku.yml` file.

Note: Migrations are executed as part of the `docker-entrypoint` script, so we don't need to run them during the release phase (just like Rails does in [this template](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/docker-entrypoint.tt#L12))

### DIFF
--- a/bin/release.sh
+++ b/bin/release.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-bundle exec rails db:migrate
-bundle exec rails feature_flags:initialize
+echo "Running release script" &&
+  bundle exec rails feature_flags:initialize
--- a/heroku.yml
+++ b/heroku.yml
@@ -11,3 +11,7 @@ run:
     command:
       - bundle exec rails jobs:work
     image: web
+release:
+  image: web
+  command:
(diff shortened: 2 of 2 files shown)


santib (approved):

bin/release.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-bundle exec rails db:migrate
-bundle exec rails feature_flags:initialize
+echo "Running release script" &&
+  bundle exec rails feature_flags:initialize
```
santib, line 4: should the db:migrate command be here as well?

JulianPasquale: We are running the migrations in the entrypoint script. Rails does the same [here](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/docker-entrypoint.tt#L12).

PerezIgnacio (approved):

PR #1003, opened 2025-03-11
Fix untracked db changes CI check

JulianPasquale:
### Description:
When we add a new migration we expect the CI to check that the `db/schema.rb` has been updated accordingly, and that's why we have a step to "Check for untracked changes in schema.rb".

As evidence you can see this [CI execution](https://github.com/rootstrap/rails_api_base/actions/runs/13791829362/job/38573448672). In this commit a new table was added. Since the CI is using the schema to setup the database, but that file was not updated, the check for untracked changes step was not failing.

By running a migrate step after loading the schema, we ensure all the differences in the schema or models annotations are applied.

Note: I'm also changing the order of the steps so the untracked changes check is performed before the assets precompile to save some time in case that fails.


santib (approved):
:nice:

PerezIgnacio (approved):

PR #1025, opened 2025-04-01
Replace DelayedJob with GoodJob

santib:
- Removed Delayed Job
- Add Good Job
  - Add Dashboard under admin auth
  - Add sample config (including cron)
  - Add heartbeat on port 3006


Procfile.dev
```diff
@@ -1,4 +1,3 @@
 web: ./bin/rails server -p 3000 -b 0.0.0.0
 js: yarn build:js --watch
 css: yarn build:css --watch
-worker: ./bin/rails jobs:work
```
JulianPasquale, line 4: > In Rails' development environment, by default, GoodJob's Adapter executes jobs async in a background thread pool in rails server

https://github.com/bensheldon/good_job?tab=readme-ov-file#set-up

I assume this is what we want for development right? Does it make sense to have it running as a standalone process? 🤔

config/initializers/good_job.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
JulianPasquale, line 1: I'm wondering if we should define some configs here, maybe creating a "high_priority" queue? No need to change now, just thinking out loud

config/environments/production.rb
```diff
@@ -57,7 +57,7 @@
   # config.cache_store = :mem_cache_store
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
-  config.active_job.queue_adapter = :delayed_job
+  config.active_job.queue_adapter = :good_job
```
JulianPasquale, line 60: I think we should enforce the `external` mode is always used in production, right? I don't think the async method is meant to be used in prod and we don't want to accidentally run it that way

```ruby
config.good_job.execution_mode = :external
```

Procfile.dev
santib: could be, I guess we need to choose between ActiveJob or GoodJob inline adapters?

config/initializers/good_job.rb
santib: ahh right, I'd go with 3 queues by default: high, default, low

config/environments/production.rb
santib: I think external is the default https://github.com/bensheldon/good_job/blob/51d90b4262070079f34615fcea9e8a8806f7de42/lib/good_job/configuration.rb#L99-L120

db/migrate/20250401200159_remove_delayed_jobs.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class RemoveDelayedJobs < ActiveRecord::Migration[8.0]
```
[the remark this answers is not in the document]
santib: hmm good point, yeah maybe we have to cleanup the migrations folder

Procfile.dev
JulianPasquale: Yeah, I think using GoodJob's default setup is ok

santib: I just pushed a commit using good_job in all envs, but now I realize I didn't get your comment. GoodJob don't need an external process for dev env because it uses a background thread within the rails server, right?

>In Rails' development environment, by default, GoodJob's Adapter executes jobs async in a background thread pool in rails server.

config/initializers/good_job.rb
santib: I was checking, and by default good job will process all jobs from all queues. Do you want to add 

`config.good_job.queues = "high;default;low"`

to limit the queues that we are going to process? We also need to take care of ActiveStorage (and others like mailers?) Rails queues I guess?

Procfile.dev
JulianPasquale: Yes, that's correct! We could still use the external process for dev if we want to, but I think the background thread is ok for development. My comment was just to confirm that 🙂

config/initializers/good_job.rb
JulianPasquale: What about
```ruby
config.good_job.queues = "high;default;low;*"
```

I think this will cover all the scenarios, but also assign more "priority" to the queues we define.

santib: what I mean is that we don't need to configure queues, it's just for the workers to decide which queues they process

santib: hmm ok but I guess we don't want to process low before others(e.g. mailers)

JulianPasquale: Agree it is not needed, just though it would be nice to have a queue that will take higher priority than a mailer for instance if we want to. But I'm ok with no providing defaults for this anyway, it can be defined if needed

santib: can you check now? I put `  config.good_job.queues = "high;*"` by default, and also enabled this option by default https://github.com/bensheldon/good_job/pull/1311 to not depend on transactions

JulianPasquale (approved):
👏

PerezIgnacio (approved):

jpascual1994 (approved):

andresg4 (approved):

PR #1044, opened 2025-04-08
Add devcontainers extending existing docker setup

JulianPasquale:
### Description:
Add settings to use devcontainers with VScode and similar IDEs (e.g. Cursor).

This config is using our existing `Dockerfile.dev` and `docker-compose.yml` files, and overriding the CMD with `sleep infinity` so it runs as an endless process.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -1,0 +1,29 @@
+services:
+  web:
+    # Uncomment if you want to override the service's Dockerfile to one in the .devcontainer 
+    # folder. Note that the path of the Dockerfile and context is relative to the *primary* 
+    # docker-compose.yml file (the first in the devcontainer.json "dockerComposeFile"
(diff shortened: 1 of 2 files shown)


.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    # Overrides default command so things don't shut down after the process ends.
+    command: sleep infinity
+
+  chrome-server:
+    image: seleniarm/standalone-chromium
+    ports:
+      - 4444:4444
+      - 7900:7900
+      - 5900:5900
```
JulianPasquale, line 28: This file is duplicating what we have in our `docker-compose.test.yml` file. I think we should try using this setup, and if people feels good about it, we can get rid of that test compose entirely in favor of devcontainers. For now, I'd rater keep both just in case

PerezIgnacio (approved):
👏

PR #1109, opened 2025-07-02
Setup New Relic dashboards and alerts with terraform

JulianPasquale:
This PR adds: 
- New Relic setup via Terraform
- Dashboard for monitoring:
  - Transactions
  - Errors
  - Key metrics
- Alerts for:
  - Application errors
  - P90 and P95 latency thresholds
  - Notifications via Slack and/or email
  - Latency configurable per-transaction


infra/new_relic/.env.sample
```diff
@@ -0,0 +1,3 @@
+NEW_RELIC_ACCOUNT_ID=1234
+NEW_RELIC_API_KEY=NRAK-**
```
santib, line 2: is this ok? in the `config/newrelic.yml` you are changing `NEW_RELIC_API_KEY` for `NEW_RELIC_LICENSE_KEY`

santib (approved):
I think we could merge it as it is and we can fix/tune it as we start using it 👌 

ty for this work

jiv (commented):
This looks great! My only concern is that there's no remote state for Terraform. All resources will be tracked using local state (created by the person who runs the terraform commands and saved locally). This means that if anyone else wants to update the NewRelic resources, Terraform won't know that there are already resources created and will create new ones.

You can configure remote state using an S3 bucket and a DynamoDB table, it's not a big deal and it's a huge win. If this is something you think it's worth it, let me know and we can discuss further.

PR #1378, opened 2026-04-20
Improve Claude Code setup

santir489:
## Description

This PR improves the Claude Code setup for `rails_api_base` by adding project-specific guidance, agents, commands, rules, and skills focused on Ruby, Rails, RSpec, PR creation, and code review workflows.

Changes include:

- Adds Claude Code skills for Rootstrap Ruby, Rails, and RSpec conventions.
- Adds a `code-reviewer` subagent to review diffs against the base branch using project conventions.
- Adds a `pr-size-checker` subagent to detect large PRs and suggest logical splits.
- Adds a `create-pr` command to standardize PR creation using the repository template.
- Adds GitHub PR formatting rules to guide consistent PR descriptions.
- Adds `CLAUDE.md` as the main entry point for Claude Code project context.
- Updates the README with Claude Code setup information.

## Notes

This is documentation/tooling only. It does not change application runtime behavior.

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -4,7 +4,7 @@ color: green
 description: "Use proactively after completing code changes, or when the user asks for a code review. Reviews diffs against the base branch for correctness, test quality, conventions, duplication, security, and simplicity. Returns structured findings."
 model: sonnet
 maxTurns: 10
-tools: Read, Grep, Glob, Bash, WebFetch
+tools: Read, Grep, Glob, Bash
 skills:
   - ruby-conventions
   - rails-conventions
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -1,0 +1,149 @@
+---
+name: code-reviewer
+color: green
+description: "Use proactively after completing code changes, or when the user asks for a code review. Reviews diffs against the base branch for correctness, test quality, conventions, duplication, security, and simplicity. Returns structured findings."
+model: sonnet
+maxTurns: 10
+tools: Read, Grep, Glob, Bash, WebFetch
+skills:
+  - ruby-conventions
+  - rails-conventions
+  - rspec-conventions
+---
+
+# Code Review Agent
+
+You review code changes. Your job is to catch issues before they become PR feedback.
+
+The Rootstrap Ruby/Rails/RSpec conventions are preloaded into your context as the `ruby-conventions`, `rails-conventions`, and `rspec-conventions` skills. **Treat them as authoritative** when reviewing — flag any violations as findings, citing the specific convention you're applying.
+
+---
+
+## Input
+
+You will receive one of:
+
+### Mode 1: File list (default)
+- A list of modified/created files
+- The original intent — either a multi-step plan summary or a simple task description
+- Read files directly from the working directory
+
+### Mode 2: Branch name
+- A branch name (e.g., `feature/add-soft-delete`)
+- Optional: original intent (if not provided, infer from commit messages)
+- Fetch both the branch and base ref to ensure accurate diffs:
+  ```bash
+  git fetch origin <branch> main
+  git diff --name-status origin/main...origin/<branch>
+  ```
+- Handle each file based on its status:
+  - **A (added) / M (modified):** Read from branch via `git show origin/<branch>:<file-path>`
+  - **D (deleted):** Read the old version via `git show origin/main:<file-path>` for context, note it was deleted in your review
+  - **R (renamed):** Read the new path from branch, note the rename in your review
+- For understanding what changed, use per-file diffs: `git diff origin/main...origin/<branch> -- <file-path>`
+- Do NOT checkout the branch
+
+### Mode 3: PR number
+- A GitHub PR number or URL
+- Fetch PR details and changed files:
+  ```bash
+  gh pr view <pr> --json title,body,headRefName,changedFiles
+  gh pr diff <pr> --name-only
+  ```
+- Extract intent from the PR title and description
+- Fetch the PR branch and base ref: `git fetch origin <head-branch> main`
+- Get file statuses: `git diff --name-status origin/main...origin/<head-branch>`
+- Handle each file by status (A/M → read from branch, D → read from `origin/main` for context, R → read new path)
+- Use per-file diffs for context: `git diff origin/main...origin/<head-branch> -- <file-path>`
+- Do NOT checkout the branch
+
+---
+
+In all modes, read all relevant files before starting your review. Use the original intent to judge whether the implementation is complete and correct.
+
+---
+
+## Review Checklist
+
+### All Code
+
+#### Logic & Correctness
+- Does the code do what the tests assert?
+- Are there logic errors the tests don't catch?
+- Edge cases handled appropriately?
+- No obvious bugs or runtime errors?
+- Error handling is appropriate?
+- No race conditions or concurrency issues?
+
+#### Test Quality
+- Are edge cases covered (nil, empty, boundary values)?
+- Do tests describe behavior, not implementation details?
+- Are test descriptions clear and accurate?
+- Any missing sad-path tests?
+- No flaky test patterns?
+- Test isolation — no order dependencies between tests
+
+#### Code Quality
+- Code is readable and follows project conventions (see preloaded skills)
+- No unnecessary complexity or over-engineering
+- Naming is clear and consistent at the design level (does the name reflect intent?)
+- No code duplication that should be extracted
+- Consistent return types
+- Single Responsibility — each class/method does one thing
+
+#### Security
+- No hardcoded secrets or credentials
+- User input is validated/sanitized
+- Authorization checks present where needed
+- No sensitive data in logs or error messages
+- No XSS vulnerabilities (unsanitized user content in HTML/JSX)
+- No CSRF vulnerabilities in state-changing endpoints
+
+#### Performance
+- No unnecessary loops or redundant computations
+- Appropriate use of caching where beneficial
+- No memory leaks or resource exhaustion risks
+
+#### Data & Queries
+- N+1 queries avoided (proper `includes`/`preload`)
+- Database-level constraints (NOT NULL, unique indexes) alongside ActiveRecord validations
+- Queries scoped to current user/tenant where needed
+- Wrap multi-step mutations in transactions
+- Database queries are efficient
+
+#### Rails Patterns
+- Strong parameters used correctly
+- Callbacks don't have side effects
+- Avoid `default_scope`
+- Background job idempotency
+- No mass-assignment vulnerabilities
+- Pundit authorization policies required for new endpoints
+- Feature flags use Flipper (see `config/initializers/flipper.rb` and the Flipper UI)
+
+#### Testing
+- Deterministic — time-dependent tests use `freeze_time` / `travel_to`
+- FactoryBot factories updated if models changed
+
+---
+
+## Output Format
+
+Return your review in this structure:
+
+```
+## Review: [Approved | Changes Requested]
+
+### Findings
+- **[severity: low/medium/high]** `file:line` — description. Suggestion: ...
+
+### Summary
+One-paragraph overall assessment.
+```
+
+### Severity scale
+
+- **high** — bugs, security issues, data loss risk, or anything that will break production. Must be fixed before merge.
+- **medium** — convention violations, missing tests for important paths, performance issues that aren't critical. Should be fixed before merge.
+- **low** — style nits, naming suggestions, optional refactors. Author can choose to address or ignore.
+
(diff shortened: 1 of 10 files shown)


mdornel (approved):

CLAUDE.md
```diff
@@ -0,0 +1,107 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What this repo is
+
+Rootstrap's Rails API template — a Rails 8.1 / Ruby 4.0 boilerplate for JSON APIs. It is meant to be cloned and renamed via `bin/bootstrap.sh --name=<project>`, which `sed`s `rails_api_base` to your project name across the codebase. The template ships with two parallel surfaces:
```
afmicc, line 7: NIT
Is there a way to ensure this section gets updated whenever a new project is started?
Maybe it doesn't belong here, and it fits better somewhere else, like in the Foundation checklist.

afmicc (approved):

.claude/skills/rails-conventions/SKILL.md
```diff
@@ -0,0 +1,154 @@
+---
+name: rails-conventions
+description: Rootstrap Rails conventions. Use when writing, reviewing, or editing any Rails code — controllers, models, migrations, routes, views, mailers, initializers, locale files, or Rails config. Covers routing, ActiveRecord, migrations, i18n, time zones, mailers, assets, Bundler groups, and logging.
+---
```
IgnacioN99, line 5: nit: Since `CLAUDE.md` mentions that project-level skills auto-activate based on the files being edited, maybe we could add `paths:` to the skill frontmatter to make that behavior explicit and deterministic.

That way, activation is driven by file glob matches rather than relying only on whether the agent considers the skill relevant from the description.

Reference: https://code.claude.com/docs/en/skills#frontmatter-reference

Example:

```yaml
---
name: ruby-conventions
description: Use when writing, reviewing, or editing Ruby files.
paths: "**/*.rb,**/*.rake,**/Gemfile,**/Rakefile,**/*.gemspec"
---
```

CLAUDE.md
```diff
@@ -0,0 +1,107 @@
+# CLAUDE.md
```
IgnacioN99, line 1: Would it be useful to include a starter `.claude/settings.json` with hooks for the workflow described in `CLAUDE.md`?

Right now, `CLAUDE.md` documents steps like loading the relevant skill, running checks, and using the reviewer agent, but those seem to rely on the agent following the instructions manually. A small hooks template could make the intended workflow easier to adopt consistently across projects.

Reference: https://code.claude.com/docs/en/hooks

For example:

```json
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write",
        "hooks": [
          {
            "type": "command",
            "command": "bundle exec rubocop",
            "timeout": 30,
            "statusMessage": "Running RuboCop..."
          }
        ]
      }
    ]
  }
}
```

Even if we do not want to enforce hooks by default, documenting this as an optional per-project setup could help clarify how teams can make the conventions more deterministic.

.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,149 @@
+---
+name: code-reviewer
+color: green
+description: "Use proactively after completing code changes, or when the user asks for a code review. Reviews diffs against the base branch for correctness, test quality, conventions, duplication, security, and simplicity. Returns structured findings."
+model: sonnet
+maxTurns: 10
+tools: Read, Grep, Glob, Bash, WebFetch
```
IgnacioN99: Would `WebFetch` be needed for the `code-reviewer` agent?

I noticed the agent declares:

```yaml
tools: Read, Grep, Glob, Bash, WebFetch
```

but I could not find anything in the agent instructions that requires fetching external URLs. Since this agent seems focused on reviewing the local codebase, would it be safer to keep the tool list limited to only what it needs?

For example:

```yaml
tools: Read, Grep, Glob, Bash
```

This would reduce the available surface area while still allowing the agent to read, search, inspect, and run local review commands. Feel free to ignore this if I’m missing a use case for `WebFetch` here.

.claude/agents/pr-size-checker.md
```diff
@@ -0,0 +1,195 @@
...
+1. **If a base branch was provided as input**, use it directly.
+2. **If there's an open PR for the current branch**, get its base:
+   ```bash
+   gh pr view --json baseRefName --jq '.baseRefName' 2>/dev/null
+   ```
+3. **If the current branch tracks an upstream**, infer from merge base:
+   ```bash
+   git merge-base HEAD main 2>/dev/null
+   git merge-base HEAD main 2>/dev/null
+   ```
```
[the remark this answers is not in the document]
IgnacioN99: Since `git merge-base` prints the merge-base SHA to stdout, running it without assigning the result does not affect the later `git diff $BASE...HEAD` call. As far as I can tell, `$BASE` still remains set to `main`.

If the intent is to handle cases where `HEAD` may not be directly comparable to `main`, we could capture the merge base instead:

```bash
BASE=$(git merge-base HEAD main 2>/dev/null || echo "main")
```

Otherwise, I think the duplicate commands can just be removed.

.claude/commands/create-pr.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+### 4. Generate PR Description
+
+Follow `.claude/rules/github-pr-formatting.md` to fill each section of the template (Board, Description, Notes, Tasks, Risk, Preview).
+
+### 5. Create PR
+
+```bash
+gh pr create --title "[{TICKET-ID}] {Concise description}" --body "{BODY}" --base main
```
IgnacioN99: nit: `create-pr.md` currently hardcodes `main` as the PR base branch:

```bash
gh pr create --title "..." --body "..." --base main
```

Since this template may be used by teams with a different default branch, such as `develop` or `master`, it might be safer to resolve the repository default branch dynamically instead of assuming `main`.

For example:

```bash
BASE_BRANCH=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name' 2>/dev/null || echo "main")

gh pr create --title "..." --body "..." --base "$BASE_BRANCH"
```

This would keep the command aligned with each repository’s actual default branch and avoid creating PRs against the wrong base by default.

.claude/skills/rspec-conventions/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+  ```
+- Avoid asserting incidental state (e.g. `Article.count == 2`) — test the delta or direct effect.
+
+## Factories & Fixtures
+- Use **FactoryBot** (`create(:article)`); **never** Rails fixtures.
+- Avoid `ModelName.create` in integration specs — reach for the factory.
+- Use **Faker** for fake data; **Database Cleaner** keeps state isolated.
+
+## Mocking / Stubbing / Doubles
+- Use mocks/stubs sparingly — favor them in isolated/behavioral specs, not integration specs.
```
PerezIgnacio, line 66: nice, I would also add something like:
-  **Avoid `allow_any_instance_of`**, as it stubs the method on every instance of the class, decoupled from which instance is actually used, and masks real interaction bugs. Prefer stubbing/mocking the specific instance. (and an example maybe)

.claude/skills/ruby-conventions/SKILL.md
```diff
@@ -0,0 +1,190 @@
...
+- Spaces around `=` in default params: `def f(a = 1)`.
+- Avoid line continuation `\` except for string concatenation.
+- In multi-line chains, keep `.` on the next line.
+- Use underscores in big numbers: `1_000_000`. Lowercase prefixes `0x`, `0o`, `0b`.
+- No block comments (`=begin`/`=end`).
+- Never more than one consecutive blank line.
+- Align multi-line method args after `(`, OR single-indent with `)` on its own line (avoid "double indent").
+- For `case`/`if` result assignment, either align branches under the keyword or break with `kind =` on its own line.
+
+## Syntax
```
PerezIgnacio, line 32: most of the items in these two sections seem to be already enforced by rubocop. should we trim the list a bit to what's not covered? mostly to reduce the context and avoid having two sources of truth (which could drift)

.claude/skills/ruby-conventions/SKILL.md
```diff
@@ -0,0 +1,190 @@
...
+- No numeric separation: `some_var1` not `some_var_1`.
+- One class/module per file, file named `snake_case` after the class/module.
+- Define non-bang methods in terms of bang when possible: `def flatten_once; dup.flatten_once!; end`.
+
+## Comments
+- Prefer self-documenting code; comments in English, capitalized, one space after `#`.
+- Refactor bad code instead of explaining it.
+- Avoid superfluous comments (`counter += 1 # Increments counter by one`).
+- Keep comments up to date — an outdated comment is worse than none.
+- Annotations: `TODO`, `FIXME`, `OPTIMIZE`, `HACK`, `REVIEW` + `:` + description, above the relevant code. Document custom annotation keywords in the project README.
```
PerezIgnacio: do we want this? I'd probably just keep TODO. Additional tags can always be added later if they are useful or a convention in the project

.claude/skills/ruby-conventions/SKILL.md
```diff
@@ -0,0 +1,190 @@
...
+- Don't nest multi-line classes; use matching folder structure.
+- Prefer modules (`extend self`) over classes with only class methods.
+- Use `def self.method` (not `def ClassName.method`).
+- Within class methods calling siblings, omit `self.`.
+- Always supply `to_s` for domain objects.
+- Use `attr_reader`/`attr_accessor`; avoid `attr`; don't prefix with `get_`/`set_`.
+- `Struct.new` for trivial value objects; don't inherit from it.
+- Avoid class variables (`@@var`); prefer class instance variables.
+- Proper visibility (`private`/`protected`); indent modifiers at method level with blank lines.
+- Prefer duck-typing over inheritance.
```
PerezIgnacio: maybe "Prefer composition over inheritance. Use duck typing where appropriate." instead?

PerezIgnacio (approved):
👏

CLAUDE.md
santir489: Good point. I will update `bin/bootstrap.sh` so the bootstrap process also replaces `rails_api_base in CLAUDE.md

.claude/agents/pr-size-checker.md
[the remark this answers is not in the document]
santir489: Good catch. I removed the duplicated git merge-base command and updated the snippet so it actually assigns the resolved merge base to BASE, which is what the later diff commands expect

.claude/skills/rails-conventions/SKILL.md
santir489: Good suggestion. I agree paths: would make skill activation more deterministic. I’d prefer to leave this as a follow-up improvement so we can define the right glob coverage for all project skills together, instead of adding it only to one skill in this PR.

CLAUDE.md
santir489: I’d keep hooks as an optional follow-up, since the right commands depend on each project’s setup and could add noise to every edit.

.claude/agents/code-reviewer.md
santir489: Agreed. Although the agent accepts a PR URL, the documented flow resolves it through gh and then reviews the fetched diff locally. There isn’t currently a WebFetch-based workflow here, so I removed WebFetch to keep the reviewer’s tool surface focused.

.claude/commands/create-pr.md
santir489: Agreed. Since this is a template, I updated the PR creation guidance to resolve the repository default branch dynamically via gh repo view, with main as a fallback, instead of hardcoding --base main.

.claude/skills/rspec-conventions/SKILL.md
santir489: Good call. I added guidance to avoid allow_any_instance_of and prefer stubbing or mocking the specific instance used by the example, since broad instance stubs can hide interaction bugs.

.claude/skills/ruby-conventions/SKILL.md
santir489: Agreed, that makes sense. I’ll leave the broader trimming as a follow-up so we can reduce RuboCop overlap intentionally and keep RuboCop as the source of truth for mechanical style.

.claude/skills/ruby-conventions/SKILL.md
santir489: Agreed. I simplified this to only recommend TODO: for now. Additional annotation tags can be added later if they become an explicit project convention.

.claude/skills/ruby-conventions/SKILL.md
santir489: Agreed, that wording is clearer. I updated it to “Prefer composition over inheritance. Use duck typing where appropriate.”

jpascual1994 (approved):

PR #1379, opened 2026-04-24
Security: Update erb gem to patch CVE-2026-41316  

IgnacioN99:
Updates erb from 6.0.2 to 6.0.4 to mitigate [CVE-2026-41316](https://www.ruby-lang.org/en/news/2026/04/21/erb-cve-2026-41316/), a vulnerability that allows bypassing  
  ERB's deserialization guard via def_method, def_module, and def_class, potentially enabling        
  arbitrary code execution when handling untrusted input.                                            
                                                                                                     
  Since erb is a transitive dependency (not declared directly in the Gemfile), it is explicitly      
  pinned to >= 6.0.4 to ensure the safe version is enforced across all environments and survives
  future lockfile regenerations.


Gemfile
```diff
@@ -15,6 +15,7 @@ gem 'cssbundling-rails', '~> 1.4'
 gem 'devise', '~> 4.9'
 gem 'devise_token_auth', '~> 1.2', '>= 1.2.6'
 gem 'draper', '~> 4.0'
+gem 'erb', '>= 6.0.4'
```
santib: I don't think we need to have erb in the Gemfile. Updating it just in the Gemfile.lock is enough

IgnacioN99: That's a fair point. The lockfile is committed and pins the version, so it does protect anyone  cloning the repo today.                                                                            
                                                                                                     
The reasoning for keeping it in the Gemfile was that the lockfile can be silently overwritten — if 
someone runs bundle update on an unrelated gem and the solver touches the erb dependency chain, it 
could resolve back to 6.0.2 with no warning. The explicit constraint prevents that regression.     
                                                                  
That said, if the team prefers to keep the Gemfile clean of transitive deps, the line can be       
removed and rely on the lockfile alone.

santib: that makes no sense. Gemfile.lock is not gonna be reverted.

santib (approved):

Gemfile.lock
```diff
@@ -670,6 +670,7 @@ DEPENDENCIES
   devise_token_auth (~> 1.2, >= 1.2.6)
   dotenv-rails (~> 3.2.0)
   draper (~> 4.0)
+  erb (>= 6.0.4)
```
santib: oops this is still wrong

santib (changes requested):
left one last comment

santib (approved):
thanks for the fix!

